Issue #214, opened 2025-12-15
Copy-paste doesn't work


sveinbjornt:
Coopying and pasting within the editor window doesn't work.

Version 3000.0.0 (1236)

macOS 15.7.2 (24G325

kojika:
me too.

App: 3000.0.0 (1236)
macOS: Tahoe 26.2 (25C56)

schuyler:
Ok, will look into this today! Might be a result of the recent changes to the pasteboard media type.

schuyler:
## Root Cause Analysis

**Issue:** Copy-paste within the editor doesn't work (reported in version 3000.0.0)

**Root Cause:** Commit `bd2e293` (Address issue #196: Add markdown UTType when copying from editor) introduced a regression.

The `writeSelectionToPasteboard:types:` override in `MPEditorView.m` completely replaces `NSTextView`'s implementation instead of extending it:

1. It calls `declareTypes:owner:` which clears the pasteboard and declares all requested types
2. But it only writes data for `NSPasteboardTypeString` and `kMPMarkdownPasteboardType`
3. If the `types` array contains other standard types (RTF, attributed strings, etc.) that NSTextView normally handles, those types get declared but **no data is written for them**

When the system requests a type that isn't handled, the paste fails because the pasteboard has no data for it.

## Fix Approach

Call `super` first to let `NSTextView` handle standard pasteboard behavior, then add our markdown type on top using `addTypes:owner:` (which doesn't clear existing pasteboard data).

## Testing

- No TDD for this regression fix
- Verify via CI that existing tests pass
- Manual testing to confirm copy-paste works within editor

schuyler:
Fixed by #220. Thank you @sveinbjornt @kojika!

Issue #196, opened 2025-11-28
Use net.daringfireball.markdown UTType when copying to pasteboard


schuyler:
When copying content from the editor, MacDown places it on the pasteboard as `public.utf8-plain-text` only. It should also include `net.daringfireball.markdown` as the UTType to allow receiving applications to identify the content as Markdown.

## Current Behavior
- Editor copy: `public.utf8-plain-text` only
- Preview copy: webarchive, RTF, HTML, plain text

## Expected Behavior
Include `net.daringfireball.markdown` (encoded as UTF-8) when copying from the editor, improving interoperability with other Markdown-aware applications.

---

Cloned from: https://github.com/MacDownApp/macdown/issues/1342

schuyler:
## Requirements Clarification

Based on discussion, here are the confirmed requirements for this implementation:

### Scope
- **Copy (⌘C) and Cut (⌘X):** Both operations will include the markdown UTType when placing content on the pasteboard
- **Drag and Drop:** Not specifically targeted, but may benefit if implementation shares code

### Pasteboard Format
- **Add:** `net.daringfireball.markdown` (UTF-8 encoded)
- **Keep:** `public.utf8-plain-text` for backward compatibility with apps that don't recognize the markdown UTType

### Behavior
- Same behavior whether copying a partial selection or full document content
- Both UTTypes will contain the same text content, just with different type identifiers

### Testing
- Will verify via CI (GitHub Actions on macOS)
- No strict TDD for this straightforward pasteboard change

PR #220, opened 2025-12-20
Address issue #214: Fix copy-paste regression in editor

schuyler:
## Summary

- Fixed a regression in `MPEditorView.m` where the `writeSelectionToPasteboard:types:` method completely replaced NSTextView's implementation instead of extending it
- The bug was introduced in commit `bd2e293` (issue #196: Add markdown UTType support)
- Copy-paste broke because the override only handled `NSPasteboardTypeString` and `kMPMarkdownPasteboardType`, ignoring other types NSTextView normally handles (RTF, attributed strings, etc.)

**The fix:** Call `super` first to let NSTextView handle standard pasteboard operations, then add the markdown type on top using `addTypes:owner:` which preserves existing pasteboard data.

## Related Issue

Related to #214

### Smoke Test (minimum)
1. **Basic copy-paste** - Copy text, paste within same document → works
2. **Cut operation** - Cut text, paste elsewhere → text removed and pasted
3. **Cross-app paste** - Copy from MacDown, paste to TextEdit → works
4. **Markdown syntax** - Copy `**bold** *italic*`, paste → syntax preserved

### Regression Comparison
- Build broken version (`bd2e293`): Copy-paste fails
- Build fixed version: Copy-paste works

### Markdown UTType Verification (issue #196)
- Copy from MacDown to a Markdown-aware app (Bear, Ulysses)
- Should recognize content as Markdown

## Review Notes

- **Code Review:** Approved - fix follows standard Cocoa pattern of calling super first then extending behavior
- **Documentation:** No updates needed
- **Architecture:** The fix properly delegates to NSTextView for standard types while preserving markdown interoperability

### DIFF
--- a/MacDown/Code/View/MPEditorView.m
+++ b/MacDown/Code/View/MPEditorView.m
@@ -193,27 +193,26 @@ NS_INLINE BOOL MPAreRectsEqual(NSRect r1, NSRect r2)
 
 /** Overridden to include markdown UTType when copying to pasteboard.
  *
- * Writes plain text and net.daringfireball.markdown types to the pasteboard
- * when requested, improving interoperability with Markdown-aware applications.
+ * Adds net.daringfireball.markdown type to the pasteboard alongside standard
+ * types, improving interoperability with Markdown-aware applications.
  * This method is called by both copy: and cut: operations.
  */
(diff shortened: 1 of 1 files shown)


schuyler:
Closes #214

Issue #11, opened 2025-11-18
Drag and drop text


schuyler:
Drag and drop text functionality stopped working in version 0.7.2, though it worked in 0.7.1.

Original issue: https://github.com/MacDownApp/macdown/issues/1133 (12 comments)

schuyler:
Closed by #240, I think

PR #240, opened 2025-12-31
Fix drag & drop

wltb:
The original author changed the default behavior of drag & drop to allow inlining JPEG images as *data:image* blocks when they are dragged into the text editor.

Unfortunately, the implementation as it was overwrote the default drag behavior completely, bit this is easily fixed with a message to _super_.

While testing this, I noticed that the code has some other quirks. It inlines all file types, not just JPEGs, and the inlining isn't undo-able; I made some comments so that someone may pick this up at some point.


schuyler:
Thanks for this patch, @wltb !

Relates to #11

schuyler:
Tests are failing due to a CI configuration issue, not the code itself.

PR #248, opened 2026-01-04
Refactor file inlining.

wltb:
This uses code from #247 and shuffles code out of MPEditorView since it began to look a bit crowded there.

Tested it, worked fine.


schuyler:
@wltb thanks for this! Looks like there are some conflicts with changes to main. You want to fix or should I take a whack at it?

wltb:
@schuyler I rebased against #247, and built, tested, and force-pushed it. Should be fine!

schuyler:
Hmm, tests are still failing in CI. I'll have to download this locally and see if I can figure out why.

wltb:
@schuyler My bad, I only tested drag & drop functionality and didn't run the unit tests. The tests for EditorView failed because I removed the called methods in the class.

I removed the EditorView test file, and added one for the new class. Fewer tests are needed now because the new class method only returns MIME types, so testing for UTIs and integration can't be done anymore.

If you want those back, I can refactor again, but I don't think the code will be likely to break.

schuyler:
Merged. Thank you!!

PR #265, opened 2026-01-10
Add documentation for the table of contents token feature.

### LINKED ISSUE #264: Documentation: Explain table of contents token
In the rendering pane of the preferences window, between the options for `Detect Jekyll front-matter` and `Render newline literally` there's an option `Detect table of contents token`.

There should be an explanation of this option in the help file, but there isn't.

This option is present in the last version of MacDown, but it is not documented there, either.

I'd expect something like:

```
### Table of Contents Token

If you include a table of contents token, which looks like `[TOC]`, then I will generate a table of contents
for the document based on the headings you've defined. This will be inserted at the place where you put
the token.
```

sbeitzel:
Fix #264 by adding a description of the table of contents token feature.

### DIFF
--- a/MacDown/Resources/help.md
+++ b/MacDown/Resources/help.md
@@ -313,6 +313,10 @@ date: 2014-06-06 20:00:00
 ---
 ```
 
+### Table of Contents Token
+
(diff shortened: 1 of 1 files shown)


schuyler:
Good catch! Thank you!!

PR #267, opened 2026-01-12
Trivial copy adjustments

limi:
Just spotted these while reading the document 😊


schuyler:
Thanks for this! I did a bit of research and it seems that "aka" is AP style and generally accepted, so I decided we can compromise and use Chicago style "a.k.a." 😄 I hope that's okay. I'm grateful for your contribution and please feel free to file any other changes you think are in order!

Issue #16, opened 2025-11-18
Does not print or generate correctly when Preview Pane is hidden


schuyler:
PDF export and printing don't include new content when preview pane is hidden.

Original issue: https://github.com/MacDownApp/macdown/issues/817 (6 comments)

schuyler:
## Requirements Clarification - Issue #16

### Problem Summary
When the Preview Pane is hidden, PDF export, HTML export, printing, and copy operations don't include newly typed content because the WebView hasn't rendered the latest markdown.

### Root Cause
- When preview is hidden, `needsHtml` returns `NO`
- Text changes don't trigger WebView re-rendering
- Export/print operations use stale WebView content

### Confirmed Scope
Fix all four affected operations:
1. **Copy HTML** (`copyHtml:`) - refactor existing fix to use general mechanism
2. **Export HTML** (`exportHtml:`) - add render deferral
3. **Export PDF** (`exportPdf:`) - add render deferral
4. **Print** (`printDocumentWithSettings:`) - add render deferral

### Implementation Approach
1. Create a general `performAfterRender:` method that:
   - Accepts a completion block
   - If preview hidden: stores handler, triggers render, waits for WebView load
   - If preview visible: executes immediately
   - Invokes handler in `webView:didFinishLoadForFrame:` after render completes

2. Replace the special-case `copying` flag with the general mechanism

3. Add new property: `@property (copy) void (^renderCompletionHandler)(void);`

### Testing Approach
- TDD where feasible (test the mechanism's queueing/execution logic)
- CI verification on macOS for full integration
- Manual testing plan for edge cases

### Estimated Change Size
~50-60 lines total:
- Add ~15-20 lines for `performAfterRender:` method
- Add 1 property, remove 1 property
- Modify 4 methods to use new pattern
- Remove ~7 lines from callback that handled `copying` flag

PR #274, opened 2026-01-14
Fix export/print when Preview Pane is hidden

schuyler:
## Summary
- Implements `performAfterRender:` mechanism to defer operations until WebView renders when preview pane is hidden
- Fixes Copy HTML, Export HTML, Export PDF, and Print operations to use current document content
- Adds 17 unit tests for the render deferral mechanism
- Refactors existing `copyHtml:` fix to use the new general mechanism

## Details
When the preview pane is hidden, `needsHtml` returns NO and the WebView doesn't render markdown changes. This caused export/print operations to output stale HTML. The fix queues operations in `renderCompletionHandlers` and triggers a render cycle, then invokes handlers when `webView:didFinishLoadForFrame:` or MathJax completion fires.

## Test plan
- [x] CI passes on macOS 14, 15, 15-intel, and 26
- [ ] Manual testing: Hide preview pane, type new content, export PDF - verify new content appears
- [ ] Manual testing: Hide preview pane, type new content, print - verify new content prints
- [ ] Manual testing: Hide preview pane, type new content, Copy HTML - verify clipboard has new content

Related to #16

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -683,7 +683,7 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
         if (printDelegate && printSelector)
         {
             NSMethodSignature *signature =
-                [NSMethodSignature methodSignatureForSelector:printSelector];
+                [printDelegate methodSignatureForSelector:printSelector];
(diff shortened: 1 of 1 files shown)


schuyler:
Closes #16

PR #286, opened 2026-01-22
Fix new document windows opening at bottom-left

mbinkhorst:
## Summary                                                                                                                                               
  - Fix new document windows opening at bottom-left screen position when no saved window position exists                                                         
  - Center the window as a fallback only when frame restoration fails, preserving saved positions for all other cases                                      
                                                                                                                                                           
  ## Solution                                                                                                                                              
  Added a fallback in `windowControllerDidLoadNib:` to center the window when no saved frame position is found. This preserves the existing behavior for:  
  - Saved documents (restore position from file URL autosave)                                                                                              
  - New documents with a saved default (restore from "Untitled" autosave)                                                                                  
                                                                                                                                                           
  Only windows with no saved position at all get centered.                                                                                                 
                                                                                                                                                           
  ## Test plan                                                                                                                                             
  - [ ] Open an existing document → restores to last saved position                                                                                        
  - [ ] Create new document, move window, close, create another → restores default position                                                                
  - [ ] Clear saved default with `defaults delete app.macdown.macdown3000 "NSWindow Frame Untitled"`, create new document → centers on screen              
  - [ ] Test on multiple monitors

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -406,6 +406,8 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
         rectString = MPRectStringForAutosaveName(kMPDefaultAutosaveName);
     if (rectString)
         [controller.window setFrameFromString:rectString];
+    else
+        [controller.window center];  // No saved position;
 
     self.highlighter =
         [[HGMarkdownHighlighter alloc] initWithTextView:self.editor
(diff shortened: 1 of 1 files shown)


mbinkhorst:
Totally forgot about that order. I will try to fix this.

schuyler:
@mbinkhorst  Thanks for the patch!

mbinkhorst:
Thank you!

PR #304, opened 2026-02-09
FileURLInlining: Use more idiomatic Objectice-C

wltb:
* custom init method instead of factory class method
* throw Exception for inherited init
* uti -> type in Tests


schuyler:
Thanks @wltb !

PR #337, opened 2026-03-03
Add zoom keyboard shortcuts (Cmd+, Cmd-, Cmd+0)

### LINKED ISSUE #335: Add Cmd+/Cmd- zoom keyboard shortcuts
## Summary

Upstream [MacDown PR #1364](https://github.com/MacDownApp/macdown/pull/1364) adds `Cmd+` / `Cmd-` keyboard shortcuts for zooming the editor and preview panes together. This addresses upstream [issue #482](https://github.com/MacDownApp/macdown/issues/482), open since 2015, which has attracted steady support from users over the years.

The feature is worth having — `Cmd+` / `Cmd-` is standard macOS zoom behavior, and users expect it. However, the upstream PR has several implementation issues that mean we should build our own cleaner version rather than incorporating it directly.

## Problems with the upstream PR

1. **Removes the `previewZoomRelativeToBaseFontSize` preference guard.** The PR deletes the early-return check in `scaleWebview`, so all users always get preview scaling — even those who explicitly disabled it in Preferences. The checkbox in `MPHtmlPreferencesViewController.xib` would still exist but do nothing.

2. **Mutates a global preference as a side effect of zoom.** The zoom actions modify `self.preferences.editorBaseFont`, which persists to `NSUserDefaults`. If a user zooms in during a presentation and quits, their base font size is permanently changed. A proper implementation should use a transient per-document zoom level, like browsers do.

3. **No `Cmd+0` reset-zoom action.** Standard macOS zoom includes a "Reset Zoom" shortcut. Without it, users have no quick way to return to their configured default.

4. **No upper bound on zoom-in.** The minimum on zoom-out is 8pt, but a user can zoom in indefinitely.

5. **No menu item validation.** The actions should disable "Zoom Out" when already at minimum size and "Zoom In" at maximum.

## Proposed implementation

- Add `zoomIn:`, `zoomOut:`, and `resetZoom:` IBAction methods to `MPDocument`
- Use a transient per-document zoom multiplier rather than permanently mutating the base font preference
- Apply zoom to both editor font size and preview via `scaleWebview`
- Add View menu items: "Zoom In" (`Cmd+`), "Zoom Out" (`Cmd-`), "Actual Size" (`Cmd+0`)
- Cap zoom range at reasonable bounds (e.g. 50%–300%)
- Keep the existing `previewZoomRelativeToBaseFontSize` preference intact and working independently
- Add `validateUserInterfaceItem:` to disable menu items at zoom limits

Related to MacDownApp/macdown#1364, MacDownApp/macdown#482

ssstonebraker:
Implements per-document zoom as requested in #335.

This PR adds standard macOS zoom shortcuts:
- Cmd+ zooms in (10% increments, max 300%)
- Cmd- zooms out (10% decrements, min 50%)
- Cmd+0 resets to actual size

Implementation:
- Uses a transient per-document zoom multiplier (not saved to preferences)
- Zoom applies to both editor and preview when the preference is enabled
- Menu items validate and disable at zoom limits
- Does not modify the base font preference
- Preserves existing Cmd+Shift+0 shortcut for split view

Fixes #335

### DIFF
--- a/MacDown/Code/Document/MPDocument.h
+++ b/MacDown/Code/Document/MPDocument.h
@@ -27,4 +27,8 @@
  */
 + (NSString *)toggleCheckboxAtIndex:(NSUInteger)index inMarkdown:(NSString *)markdown;
 
+- (IBAction)zoomIn:(id)sender;
+- (IBAction)zoomOut:(id)sender;
+- (IBAction)resetZoom:(id)sender;
+
 @end
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -37,6 +37,9 @@
 
 static NSString * const kMPDefaultAutosaveName = @"Untitled";
 
+static const CGFloat kMPMinZoom = 0.5;
+static const CGFloat kMPMaxZoom = 3.0;
+
 
 NS_INLINE NSString *MPEditorPreferenceKeyWithValueKey(NSString *key)
 {
@@ -782,17 +785,15 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
     // Zoom menu validation
     if (action == @selector(zoomIn:))
     {
-        static const CGFloat kMaxZoom = 3.0;
-        return self.zoomMultiplier < kMaxZoom;
+        return self.zoomMultiplier < kMPMaxZoom;
     }
     else if (action == @selector(zoomOut:))
     {
-        static const CGFloat kMinZoom = 0.5;
-        return self.zoomMultiplier > kMinZoom;
+        return self.zoomMultiplier > kMPMinZoom;
     }
     else if (action == @selector(resetZoom:))
     {
-        return self.zoomMultiplier != 1.0;
+        return fabs(self.zoomMultiplier - 1.0) > 0.001;
     }
     else if (action == @selector(toggleToolbar:))
     {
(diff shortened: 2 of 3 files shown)


ssstonebraker:
hey sorry there is a bug i am going to update again shortly

schuyler (changes requested):
Thanks for tackling #335 — the overall design is clean. Transient per-document zoom via a multiplier (not mutating preferences) is the right call. A few things to address before this is ready to merge:

## Bug: Preview zoom broken when preference is off

This is the main issue. `scaleWebview` has an early return when `previewZoomRelativeToBaseFontSize` is `NO`:

```objc
if (!self.preferences.previewZoomRelativeToBaseFontSize)
    return;
```

Because the `zoomMultiplier` is applied *inside* that method, Cmd+/Cmd- won't affect the preview at all when the preference is off. The zoom shortcuts should work on the preview regardless of that setting.

## Duplicated zoom constants

`kMaxZoom` (3.0) and `kMinZoom` (0.5) are each defined as `static const` in three separate places — `validateUserInterfaceItem:`, `zoomIn:`, and `zoomOut:`. These should be file-level constants defined once to avoid the risk of them getting out of sync.

## Floating-point comparison in `resetZoom:` validation

```objc
return self.zoomMultiplier != 1.0;
```

After several `+0.1` / `-0.1` round trips, IEEE 754 can drift — `1.0 + 0.1 - 0.1` may not exactly equal `1.0`. This could leave the "Actual Size" menu item enabled at 100%. Something like `fabs(self.zoomMultiplier - 1.0) > 0.001` would be safer.

ssstonebraker:
Thanks for the review. Pushed a fix for all three.

Moved the zoom constants to file scope with the `kMP` prefix to match the rest of the codebase. Restructured `scaleWebview` so the zoom multiplier always applies to the preview regardless of the `previewZoomRelativeToBaseFontSize` setting. Fixed the floating-point comparison in `resetZoom:` validation.

Tested locally, everything works as expected.

schuyler:
Thank you, @ssstonebraker — this PR was the foundation for the entire zoom effort, and it's appreciated. 🙏

The three fixes you pushed in response to review (file-scope zoom constants, preview zoom working regardless of the `previewZoomRelativeToBaseFontSize` setting, and the floating-point comparison in the reset-zoom validation) are already productized in #395, with your original commits cherry-picked and credited there.

We've now consolidated the design across all the zoom PRs into a single plan in #470, which unifies editor **and** preview zoom behind one ⌘+/⌘−/⌘0 key pair. Closing this as superseded — but the lineage traces straight back to your work here. Thanks again!

---

PR #339, opened 2026-03-07
Add 'Start in preview mode' preference

cyclecraze:
Would you kindly accept this PR that adds a preference `Start in preview mode`. Thank you.

## Summary

- Adds a **Start in preview mode** checkbox to General Preferences → Behavior
- When enabled, every new window opens with the editor pane hidden, showing only the preview
- The preference is persisted via `NSUserDefaults` and applies on each launch

## Changes

**`MacDown/Code/Preferences/MPPreferences.h` / `.m`**
- Added `layoutStartViewerOnly` BOOL property with `@dynamic` so PAPreferences routes it through NSUserDefaults

**`MacDown/Code/Document/MPDocument.m`**
- In `windowControllerDidLoadNib`, after all setup completes, dispatches `toggleSplitterCollapsingEditorPane:YES` via an extra `dispatch_async` to ensure Auto Layout has settled before the pane is collapsed

**`MacDown/Localization/Base.lproj/MPGeneralPreferencesViewController.xib`**
- Added "Start in preview mode" checkbox to the Behavior box, bound to `self.preferences.layoutStartViewerOnly`
- Extended box height to accommodate the new row


schuyler (dismissed):
## PR Review: Add "Start in preview mode" preference

This is a nice feature addition — the concept is sound and the PR description is thorough. However, there are several issues that should be addressed before merging.

### Issues to Fix

**1. Property naming breaks existing conventions**

All existing preference properties follow a consistent prefix-based naming scheme: `editor*`, `html*`, `extension*`, etc. The new property `layoutStartViewerOnly` uses a `layout` prefix that doesn't exist anywhere else in the codebase. A name consistent with the existing pattern would be something like `editorStartInPreviewMode` — which is clearer in intent and follows the `editor*` prefix used by other editor behavior properties.

**2. Property placement in the header is inconsistent**

In `MPPreferences.h`, the new `layoutStartViewerOnly` property is inserted between `editorAutoSave` and `editorScrollsPastEnd`, breaking the existing grouping of `editor*` properties. The corresponding `@dynamic` declaration in `MPPreferences.m` is similarly misplaced. It should be placed with the other general behavior properties near the top (alongside `supressesUntitledDocumentOnLaunch` and `createFileForLinkTarget`), or given a proper `editor*` prefix and kept in the editor group.

**3. Double `dispatch_async` is fragile**

The implementation nests a `dispatch_async` inside the existing `NSOperationQueue mainQueue` block:

```objc
[[NSOperationQueue mainQueue] addOperationWithBlock:^{
    [self setupEditor:nil];
    [self redrawDivider];
    [self reloadFromLoadedString];
    if (self.preferences.layoutStartViewerOnly) {
        dispatch_async(dispatch_get_main_queue(), ^{
            [self toggleSplitterCollapsingEditorPane:YES];
        });
    }
    [self startFileWatching];
}];
```

This relies on the assumption that a single run-loop deferral will always be enough for Auto Layout to settle. This is a timing-dependent pattern that may not hold on all macOS versions or hardware. A more robust approach would be to use `dispatch_after` with a small delay, or better yet, use `viewDidLayout`/`updateConstraints` to detect when layout is actually complete. At minimum, a comment explaining *why* the double-dispatch is needed would help future maintainers.

**4. Inserted before `startFileWatching` — ordering concern**

The new block is placed before `[self startFileWatching]` in the source, but because it dispatches asynchronously, `startFileWatching` actually runs first. This isn't a bug, but the code reads as if the pane collapse happens before file watching starts. Consider moving the check *after* `startFileWatching` for readability, since the async dispatch means ordering doesn't matter functionally.

**5. `Podfile.lock` change is unrelated**

The `Podfile.lock` checksum changed without any corresponding change to the `Podfile`. This suggests the contributor ran `pod install` with a slightly different CocoaPods environment. This diff should be reverted to keep the lock file consistent.

**6. `changes.md` should not be committed**

The PR adds a new `changes.md` file at the repo root. This project doesn't use a `changes.md` file — there's no existing changelog convention in the repo. This file duplicates what's already in the PR description and commit message. It should be removed.

**7. XIB constraint IDs look hand-edited**

The new constraint IDs in the XIB (`cW2-nv-iRz`, `cS1-nv-iRz`, `bS1-nv-iRz`, `lS2-nv-iRz`, `tS2-nv-iRz`) share a suspiciously similar suffix pattern (`-nv-iRz`), which suggests they were hand-edited rather than generated by Interface Builder. While this *works*, hand-crafted IDs risk colliding with IB-generated IDs in future edits. The XIB changes should ideally be made through Interface Builder to get proper auto-generated IDs.

### Minor Nits

- The height constraint on the new checkbox (`constant="14"`) is hardcoded. Other checkboxes in the same XIB (e.g., `aSv-Qd-7kR`) also have height constraints, so this is consistent, but it's worth verifying this renders correctly across different macOS text size settings.

### Summary

The feature itself is well-motivated and the implementation approach (reusing `toggleSplitterCollapsingEditorPane:`) is correct. The main asks are:
1. Rename the property to follow existing naming conventions
2. Remove the extraneous `Podfile.lock` change and `changes.md` file
3. Add a comment explaining the double-dispatch timing requirement
4. Ideally regenerate the XIB changes through Interface Builder

marceloceccon:
This would be great, I was about to fork repo to add exactly this feature. 

@cyclecraze Do you plan on going ahead with this and fix those issues?

I've been on the hunt for a good MD viewer for macOS (I don't want to open on cursor/code) and pretty much other OS alternatives are outdated.

yusufm:
> This would be great, I was about to fork repo to add exactly this feature.
> 
> @cyclecraze Do you plan on going ahead with this and fix those issues?
> 
> I've been on the hunt for a good MD viewer for macOS (I don't want to open on cursor/code) and pretty much other OS alternatives are outdated.

@marceloceccon this is fixed in #383

schuyler:
And will be released in 3000.0.7!

schuyler:
@cyclecraze thank you very much for submitting this PR. Is it okay if we close it out now that the feature is implemented?

cyclecraze:
Certainly.Sorry I could not do more.Florin @Renovo.us404-919-1919On Apr 19, 2026, at 6:41 PM, Schuyler Erle ***@***.***> wrote:﻿schuyler left a comment (schuyler/macdown3000#339)
@cyclecraze thank you very much for submitting this PR. Is it okay if we close it out now that the feature is implemented?

schuyler:
Not at all! I'm grateful for the contribution. Thanks!

Issue #290, opened 2026-01-24
Request: Command+R to reload current file


btsai:
Hi - I'm not an Objective-C guy, so just wondering out loud, how hard would it be for someone who does know Objective-C to generate a PR to make the currently opened file reload when you hit Cmd+R?

With Claude Code generating documentation in Markdown, the files often get updated and the only way to see the changes is to close the file and reopen - it would be so much more productive to just press Cmd+R.

Just a thought....

Thanks for reading !

schuyler:
Would it be better or worse to just have the file reload when it changes on
disk?

btsai:
Live reload via fswatch? OMG even better! Like magic!

btsai:
<img width="578" height="411" alt="Image" src="https://github.com/user-attachments/assets/2b9eaee9-9a69-4e6b-bc15-0047e5797139" />

Paypal isn't letting me thank you for letting me use Macdown 3000....

schuyler:
I'll take the thought for the deed! Marking this as an enhancement for the next release -- the upcoming one is going to be some urgent bug fixes.

schuyler:
related to #110

schuyler:
## Requirements Clarification - fswatch-style Auto-Reload

After discussion, here are the confirmed requirements for implementing automatic file reload when the file is modified externally:

### Behavior
- **When document has no unsaved changes:** Automatically reload the file silently
- **When document has unsaved changes:** Show a prompt asking the user whether to reload (losing their changes) or keep their version

### Implementation Approach
- Use `dispatch_source` (GCD file monitoring) for efficient, event-driven file change detection
- Monitor the document's file for `DISPATCH_VNODE_WRITE` events
- Compare `fileModificationDate` to detect actual changes vs. self-triggered saves

### Preferences
- **No preference toggle** - This will be the default behavior
- The app is meant to be opinionated and easy to use
- The prompt-when-dirty behavior provides user control where it matters

### Technical Notes
- MacDown uses `autosavesInPlace` (YES), so macOS handles periodic autosaving
- Need to handle the case where our own save triggers the file watcher
- Should stop monitoring when document closes to prevent leaks

This approach addresses the original request for Cmd+R reload by making it automatic - the file reloads when changed externally without requiring any user action.

btsai:
So exciting ! 
Can't wait till this makes it to a release !
Thank you so much!

tikiverse:
thank you for addressing this!

PR #354, opened 2026-04-03
Fix file watcher dropping after atomic save by external editors

BenCello:
## Summary

- Editors like Sublime Text, VSCode, vim, and nano use **atomic saves**: they write to a temp file then `rename()` over the original
- This causes the VNODE watcher to receive a `DELETE`/`RENAME` event and **stop watching entirely**, so subsequent external changes go undetected
- The cancel handler now schedules a 0.3s restart, re-attaches to the new inode, and checks for a content change

## Guards against spurious restarts

| Check | Prevents |
|---|---|
| `weakSelf` nil | Restart after document is closed |
| `fileURL.path != path` | Restart after Save As to a different path |
| `fileWatcher.isWatching` | Double-restart (startFileWatching calls stopFileWatching internally) |
| File doesn't exist | Restart after a genuine deletion |

## Test plan

- [ ] Open a file in MacDown, edit and save from Sublime Text → silent reload
- [ ] Open a file in MacDown, edit and save from VSCode → silent reload
- [ ] Open a file, make unsaved edits, then save externally → prompt appears
- [ ] Save As to a new file, edit the new file externally → reloads correctly
- [ ] Close document while watcher is active → no crash or spurious reload

## Related

Related to #290

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -2922,7 +2922,20 @@ current file somewhere to enable this feature.", \
                  [weakSelf handleExternalFileChange];
              }
        cancelHandler:^(NSString *path) {
-                 // File was deleted or renamed — watcher auto-stopped
+                 // File was deleted or renamed (e.g. atomic save by external editor).
+                 // Wait briefly for the rename to complete, then restart watching
+                 // the new inode at the same path.
+                 dispatch_after(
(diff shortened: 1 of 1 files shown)


schuyler:
Great catch. Thank you for the PR!

PR #383, opened 2026-04-19
[codex] Persist preview-only startup mode

yusufm:
## Summary

This adds a General preference to always start documents in preview-only mode and applies it during document startup after the split view restores its autosaved layout.

## Why

MacDown already lets users hide the editor pane for a preview-only view, but that state does not survive app restarts. The root cause is that startup restores the split view's autosaved geometry and never reapplies a durable user preference for preview-only startup.

## User impact

Users who prefer MacDown as a Markdown viewer can enable `Start in preview mode` in General settings and consistently reopen documents with the editor pane hidden.

## Implementation notes

- added a persisted `editorStartInPreviewMode` preference using the existing `editor*` naming convention
- exposed the setting in the General preferences pane
- applied the preference after initial split-view autosave restoration and window layout so it reliably overrides stale editor-visible startup geometry
- preserved a sane restore ratio when startup begins from an editor-only autosaved layout
- added preference and pane-toggle tests for the new behavior

## Validation

- `ibtool --errors --warnings --notices --compile /tmp/MPGeneralPreferencesViewController.nib MacDown/Localization/Base.lproj/MPGeneralPreferencesViewController.xib`
- attempted `xcodebuild test -workspace 'MacDown 3000.xcworkspace' -scheme 'MacDown' -configuration Debug -destination 'platform=macOS' -only-testing:MacDownTests/MPPreferencesTests/testStartInPreviewModeToggle -only-testing:MacDownTests/MPPaneToggleTests/testStartInPreviewModeRestoresPreviewFromEditorOnlyLayout`
- `xcodebuild` is currently blocked in this checkout because `Pods/Target Support Files/...xcconfig` files are missing and local Bundler/CocoaPods are not usable

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -266,6 +266,7 @@ typedef NS_ENUM(NSUInteger, MPScrollOwner) {
 - (void)refreshHeaderCacheAfterResize;
 - (void)windowDidEndLiveResize:(NSNotification *)notification;
 - (void)windowDidChangeFullScreen:(NSNotification *)notification;
+- (void)applyEditorStartInPreviewModePreference;
 // Commit 8 (gap 9): MathJax generation counter accessor (used by tests via category)
 - (NSUInteger)mathJaxRenderGeneration;
 
@@ -565,6 +566,11 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
 
         // Issue #290: Start file watching for auto-reload
         [self startFileWatching];
+
+        // Apply the startup pane preference after split-view autosave and the
+        // initial window layout have produced stable geometry.
+        [controller.window.contentView layoutSubtreeIfNeeded];
+        [self applyEditorStartInPreviewModePreference];
 
         // Commit 6 (gaps 1+3): Register for window resize/fullscreen notifications.
         // Registered here (not in the main setup block) because self.editor.window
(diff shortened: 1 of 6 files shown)


schuyler:
Thank you so much for this PR! Here is some targeted feedback from Claude Code. I'd be delighted to merge if we can address these points.

Targeted feedback:

1. **`applyEditorStartInPreviewModePreference` — unreachable-looking branch.** Because of the `!self.editorVisible` early return, `dividerLocation` will only land on exactly `0.0`/`1.0` in the editor-only autosave case, so the `else if (self.previousSplitRatio < 0.0)` branch reads as defensive-but-unreachable. Please either add a one-line comment explaining that this fallback is specifically for the editor-only autosaved layout, or restructure so the intent is explicit.

2. **Test fragility in `testStartInPreviewModeRestoresPreviewFromEditorOnlyLayout`.** The assertion `previousSplitRatio == 0.5` relies on `MPDocumentSplitView.dividerLocation` returning exactly `1.0` for a 399/0 split in a 400-px view. That's an implementation detail of how ratio is computed relative to divider thickness; a small tweak to the split view could flip this test to the "save current ratio" branch silently (setting `previousSplitRatio ≈ 0.9975`) and still pass nothing. Prefer asserting observable outcomes — `editorVisible == NO`, `previewVisible == YES`, and the resulting `splitView.dividerLocation` — rather than pinning the `0.5` fallback constant.

3. **XIB default state.** The new checkbox has `state="on"` in `MPGeneralPreferencesViewController.xib` (`aYQ-0c-EOQ`). The preference should be opt-in per the PR description, and for a first-run user (no default written yet) this risks a brief visual mismatch before bindings resolve. Please set `state="off"` in the xib.

4. **Trailing newline at EOF of `MPDocument.m` was removed.** Probably unintentional — please restore it.

5. **Optional: drop the forced `layoutSubtreeIfNeeded`.** If `MPDocumentSplitView.dividerLocation` already derives from subview frames that AppKit has set by the time `windowControllerDidLoadNib:` runs, the forced layout pass is unnecessary startup work. If it's required for correctness, a comment saying *why* would help.

---

schuyler:
Looks good. Thank you so much!

PR #385, opened 2026-04-19
[codex] Harden Quick Look preview sandbox

yusufm:
## Summary

- disable content JavaScript in the Quick Look WKWebView and cancel non-local navigations
- add a restrictive Content Security Policy to Quick Look HTML output
- stop embedding Prism scripts in Quick Look and remove the extension network-client entitlement
- add renderer tests covering the CSP and the absence of embedded Quick Look scripts

## Why

Quick Look was rendering attacker-controlled Markdown into a live WKWebView with script execution paths and outbound network capability. A malicious file previewed in Finder could execute JavaScript and exfiltrate document content without opening the main app.

## Root cause

The Quick Look path reused the normal rendered HTML body, appended executable Prism scripts, left WebKit JavaScript enabled, and shipped a sandbox entitlement for outbound network access.

## Validation

- `xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme MacDown -destination "platform=macOS"`
- 257 tests passed, 0 failures

### DIFF
--- a/MacDownCore/MPQuickLookRenderer.m
+++ b/MacDownCore/MPQuickLookRenderer.m
@@ -163,6 +163,21 @@ NS_INLINE NSString *MPPreprocessMarkdown(NSString *text)
     return result;
 }
 
+NS_INLINE NSString *MPQuickLookContentSecurityPolicy(void)
+{
+    return @"default-src 'none'; "
+           @"base-uri 'none'; "
+           @"form-action 'none'; "
+           @"object-src 'none'; "
+           @"frame-src 'none'; "
+           @"connect-src 'none'; "
+           @"img-src data: file:; "
+           @"media-src data: file:; "
+           @"font-src data: file:; "
+           @"style-src 'unsafe-inline'; "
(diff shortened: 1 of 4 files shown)


schuyler:
Nice layered hardening — CSP + disabled JS + nav-cancel + entitlement drop is solid defense in depth, and I like that tests cover both the CSP presence and the absence of embedded Prism scripts.

A few non-blocking observations:

- `-[MPQuickLookRenderer embeddedScripts]` (MPQuickLookRenderer.m:419) and any now-unused Prism helpers appear to be dead code after this change. Worth pruning in a follow-up so the attack surface stays trimmed.
- `config.preferences.javaScriptEnabled = NO` is deprecated on macOS 11+. Correct for backward compat, but may produce a build warning; an `@available` else-branch or a pragma would keep the build clean.
- Allowing `data:` in `decidePolicyForNavigationAction:` is safe given the CSP and JS-off posture, but a one-line comment explaining why it's allowed would help future readers.

None of these are blockers — LGTM.

---

Issue #394, opened 2026-04-21
Crash in selectedToolbarItemGroupItem: when toolbar item identifier is missing from dictionary


schuyler:
In `MPToolbarController.m`, `selectedToolbarItemGroupItem:` does a dictionary lookup for the sender's identifier but has no guard against a nil result:

```objc
NSToolbarItemGroup *selectedGroup = self->toolbarItemIdentifierObjectDictionary[sender.identifier];
NSToolbarItem *selectedItem = selectedGroup.subitems[selectedIndex];
```

If `sender.identifier` is not present in `toolbarItemIdentifierObjectDictionary`, `selectedGroup` is `nil`, and the subsequent `subitems` call returns `nil`. Calling `objectAtIndex:` on the resulting `nil` array will raise `NSRangeException` and crash.

This is a pre-existing gap, separate from the IMP-cast crash fixed in #393. It would be triggered by any toolbar customization that produces a segmented control whose identifier the dictionary doesn't recognize.

**Fix:** add a nil guard on `selectedGroup` before accessing `.subitems`, and log/assert in the unexpected case.

schuyler:
## Requirements

**Fix scope:** `selectedToolbarItemGroupItem:` in `MPToolbarController.m`

**Guards to add:**
1. Nil guard on `selectedGroup` after the dictionary lookup — assert (NSAssert) in debug, silent return in release
2. Bounds guard on `selectedIndex` vs `selectedGroup.subitems.count` before accessing the subitem

**Behavior on guard fire:** `NSAssert` (crashes Debug builds loudly, no-ops in Release)

**Testing:** TDD — failing tests written first, then implementation. Tests will cover:
- Nil group (sender identifier not in dictionary)
- Out-of-bounds index
- Happy path (valid group, valid index, action dispatched)

**Approach:** Add tests to existing `MPToolbarControllerTests.m`, which has the controller harness already set up. No document mock needed for these tests — the guard fires before the `performSelector:` call.

---
_Generated by [Claude Code](https://claude.ai/code)_

PR #396, opened 2026-04-21
Guard nil-group and out-of-bounds index in selectedToolbarItemGroupItem:

schuyler:
## Summary

Adds nil-group and out-of-bounds guards to `selectedToolbarItemGroupItem:` in `MPToolbarController.m`.

The method previously did `selectedGroup.subitems[selectedIndex]` with no protection against:
- `sender.identifier` not being present in `toolbarItemIdentifierObjectDictionary` (`selectedGroup` becomes nil)
- `selectedIndex` being negative or `>= selectedGroup.subitems.count` (NSRangeException from `objectAtIndexedSubscript:`)

The bounds case is a latent `NSRangeException` crash. The nil case is benign today thanks to ObjC nil-messaging semantics, but is a correctness hazard if the downstream guards ever change.

Both paths now use `NSAssert` + an early `return`:
- In Debug (CI default, `ENABLE_NS_ASSERTIONS = YES`) the asserts fire loudly with a message that names the offending identifier and the actual vs expected subitem count, so misconfigurations are diagnosed immediately.
- In Release (`ENABLE_NS_ASSERTIONS = NO`) the asserts compile away and the `if (…) return;` keeps users out of the crash path.

## Test coverage

Four new tests in `MPToolbarControllerTests.m`:
- `testSelectedToolbarItemGroupItemUnknownIdentifierAsserts`
- `testSelectedToolbarItemGroupItemOutOfBoundsIndexAsserts`
- `testSelectedToolbarItemGroupItemNegativeIndexAsserts` (with a precondition check that `NSSegmentedControl` actually stores `-1`)
- `testSelectedToolbarItemGroupItemValidCallDoesNotCrash` (regression guard for the happy path)

The three guard tests use `XCTAssertThrowsSpecificNamed(…, NSInternalInconsistencyException)` because the test target is built in Debug. A private category declaration exposes `selectedToolbarItemGroupItem:` to the test target so the call compiles without `-Wundeclared-selector`.

The Release-mode `if … return;` branches are defense-in-depth and not directly reachable under Debug; this is noted in a short comment above the new test section.

## Related Issue

Related to #394

## Manual Testing Plan

N/A — the XCTest suite covers the testable surface, and the guards only fire on a misconfigured toolbar that can't be produced through the normal UI (segmented controls are constructed in `toolbarItemGroupWithIdentifier:` with matching identifier and `segmentCount == items.count`, so the fault conditions are structurally impossible from user interaction).

### DIFF
--- a/MacDown/Code/Application/MPToolbarController.m
+++ b/MacDown/Code/Application/MPToolbarController.m
@@ -111,6 +111,19 @@ static CGFloat itemWidth = 37;
     NSInteger selectedIndex = sender.selectedSegment;
 
     NSToolbarItemGroup *selectedGroup = self->toolbarItemIdentifierObjectDictionary[sender.identifier];
+    NSAssert(selectedGroup != nil,
+             @"selectedToolbarItemGroupItem: sender identifier '%@' is not registered "
+             @"in toolbarItemIdentifierObjectDictionary",
+             sender.identifier);
+    if (!selectedGroup) { return; }
+
+    NSAssert(selectedIndex >= 0 && (NSUInteger)selectedIndex < selectedGroup.subitems.count,
+             @"selectedToolbarItemGroupItem: selectedSegment %ld is out of bounds "
+             @"for group '%@' with %lu subitems",
+             (long)selectedIndex, sender.identifier,
+             (unsigned long)selectedGroup.subitems.count);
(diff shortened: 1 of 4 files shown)


schuyler:
thanks @yusufm as we caught this one while reviewing #393

PR #401, opened 2026-04-22
Fix Xcode runtime warnings

yusufm:
Removes deprecated NSToolbarItem.maxSize, adds secure restorable state opt-in, and silences misleading file:// navigation security log for untitled documents.


schuyler:
Thank you @yusufm !

PR #409, opened 2026-05-03
Harden preview script CSP

yusufm:
## Summary
- add per-render nonces to renderer-owned preview script tags
- remove broad file/self script sources and outbound connect permission from the preview CSP
- add regression coverage for nonce-bearing app scripts and document-supplied file scripts

## Validation
- bundle exec pod install
- xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme MacDown -only-testing:MacDownTests/MPAssetTests -only-testing:MacDownTests/MPRendererEdgeCaseTests/testPreviewRenderIncludesContentSecurityPolicyAndCheckboxToken -only-testing:MacDownTests/MPRendererEdgeCaseTests/testPreviewCSPBlocksDocumentSuppliedFileScripts -destination 'platform=macOS'

Related to security review finding: Preview CSP allows local script execution.

### DIFF
--- a/MacDown/Code/Document/MPAsset.h
+++ b/MacDown/Code/Document/MPAsset.h
@@ -35,6 +35,7 @@ extern NSString * const kMPMathJaxConfigType;
 
 @interface MPScript : MPAsset
 + (instancetype)javaScriptWithURL:(NSURL *)url;
+- (NSString *)htmlForOption:(MPAssetOption)option nonce:(NSString *)nonce;
 @end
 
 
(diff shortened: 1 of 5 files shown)


schuyler (changes requested):
Thanks for the PR, @yusufm!

## Issues

- **MathJax/preview network access:** Tightening `connect-src` from `http: https:` to `'none'` blocks all XHR/fetch from the preview, so it's worth confirming MathJax rendering and the other bundled preview libraries still work, since the prior policy granted that access explicitly.

## Suggestions

- **Nonce robustness:** The nonce is injected into the script attribute without escaping — safe today since it's an NSUUID hex string, but a small assertion or escape would keep it sound if the generator ever changes.
- **Pin the nonce contract in tests:** A test asserting the CSP `nonce-…` value equals the nonce on the emitted renderer scripts (and that it differs across renders) would lock in the core security property.
- **Changelog:** A `### Security` entry under `[Unreleased]` would record that document-supplied `file://` preview scripts are now blocked.

---

Issue #378, opened 2026-04-18
External file change temporarily corrupts editor syntax highlighting


schuyler:
## Description

When a file is modified externally while open in MacDown, the editor's syntax highlighting can become corrupted. All text following the cursor position at the time of the external change renders in the wrong style (e.g., h2 cyan color and font size). The corruption self-corrects after scrolling.

## Steps to Reproduce

1. Open a document in MacDown
2. Position the cursor partway through the document
3. Externally append a header (e.g., `echo "## h2" >> file.md`)
4. Observe the editor

## Expected Behavior

Editor re-highlights correctly after the file revert.

## Actual Behavior

Text after the cursor renders in incorrect styling (cyan h2 style observed). Self-corrects after scrolling up and down.

## Likely Cause

The syntax highlighter (`HGMarkdownHighlighter`) may not fully re-highlight on the revert path when text is replaced via `reloadFromLoadedString`.

Related to #342

PR #415, opened 2026-05-03
Clear editor highlighting on reload

yusufm:
## Summary
- clear existing editor highlighting after replacing text during reload from disk
- refresh clear-text attributes before scheduling the new highlight pass
- add a lifecycle regression for the external reload path

Related to #378.

## Validation
- xcodebuild test -workspace 'MacDown 3000.xcworkspace' -scheme MacDown -destination 'platform=macOS' -only-testing:MacDownTests/MPDocumentLifecycleTests/testExistingDocumentClearsHighlightingBeforeReloadHighlight -only-testing:MacDownTests/MPDocumentLifecycleTests/testExistingDocumentTriggersRenderOnReload -only-testing:MacDownTests/MPDocumentLifecycleTests/testExistingDocumentTriggersHighlightOnReload

Note: the last only-testing selector is not present on main, so Xcode ran the two matching lifecycle tests.

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -594,6 +594,8 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
         {
             self.editor.string = self.loadedString;
             self.loadedString = nil;
+            [self.highlighter clearHighlighting];
+            [self.highlighter readClearTextStylesFromTextView];
         }
 
         // Gap 8: Claim editor ownership before rendering so that the full-reload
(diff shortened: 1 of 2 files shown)


schuyler (approved):
Looks great — thanks @yusufm. The `clearHighlighting` → `readClearTextStylesFromTextView` → `parseAndHighlightNow` sequence correctly mirrors the highlighter's own ordering in `applyStylesFromStylesheet:`, and since `parseAndHighlightNow` only schedules an async parse, the explicit synchronous clear is what actually closes the stale-attribute gap from #378. Nice regression test too. Approving.

---

Issue #291, opened 2026-01-26
Feature suggestion: add support for Github alerts formating


5T33Z0:
As you may know, Github markdown supports alerts which help drawing the reader's attention: 

> [!NOTE]  
> Highlights information that users should take into account, even when skimming.

Source:
```markdown
> [!NOTE]  
> Highlights information that users should take into account, even when skimming.
```

> [!TIP]
> Optional information to help a user be more successful.

Source:
```markdown
> [!TIP]
> Optional information to help a user be more successful.
```

> [!IMPORTANT]  
> Crucial information necessary for users to succeed.

Source:
```markdown
> [!IMPORTANT]  
> Crucial information necessary for users to succeed.
```

> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.

Source:
```markdown
> [!WARNING]  
> Critical content demanding immediate user attention due to potential risks.
```

> [!CAUTION]
> Negative potential consequences of an action.

Source:
```markdown
> [!CAUTION]
> Negative potential consequences of an action.
```

I wonder if this could be implemented. Or is this part of the CSS file when rendering?

I am using `GitHub2` as the CSS file and with it, a Note block looks like this when rendered:

<img width="403" height="97" alt="Image" src="https://github.com/user-attachments/assets/73b8d974-dd0e-47ca-8269-a424ce264951" />

Basically, it just looks like a regular quote and ignores the `[!NOTE]` "trigger" so to speak.

schuyler:
I wholeheartedly approve. However, this may be blocked on replacing Hoedown with something more modern (#77) which is liable to be a big and somewhat risky project. I'm happy to leave this open but I can't promise when we'll get it implemented. I definitely would welcome help here.

5T33Z0:
Ah, okay. Before you took on the challenge to make MacDown usable on current macOS versions, I looked around for other options and ended up with VSCodium and some [plugins](https://marketplace.visualstudio.com/publishers/bierner) by Matt Bierner to handle the Github Flavoured Markdown. But I don't know which parser it uses.

schuyler:
GitHub alerts (aka admonitions) are supported natively by cmark-gfm. This should be straightforward to implement once #77 (migrate from Hoedown to a CommonMark parser) lands in v3000.1.x.

PR #418, opened 2026-05-03
Render GitHub alert blockquotes

yusufm:
Related to #291

## Summary
- Convert GitHub-style blockquote alerts into semantic alert containers after Markdown rendering
- Add bundled alert CSS for preview and styled HTML export
- Add renderer tests for alert conversion and ordinary blockquotes

## Tests
- xcodebuild test -workspace 'MacDown 3000.xcworkspace' -scheme MacDown -destination 'platform=macOS' -only-testing:MacDownTests/MPRendererEdgeCaseTests/testRendererFormatsGitHubAlertBlockquote -only-testing:MacDownTests/MPRendererEdgeCaseTests/testRendererLeavesNormalBlockquoteAlone -only-testing:MacDownTests/MPRendererEdgeCaseTests/testPreviewRenderIncludesContentSecurityPolicyAndCheckboxToken

### DIFF
--- a/MacDown/Code/Document/MPRenderer.m
+++ b/MacDown/Code/Document/MPRenderer.m
@@ -192,6 +192,59 @@ NS_INLINE NSString *MPPreprocessMarkdown(NSString *text)
     return mut;
 }
 
+NS_INLINE NSString *MPHTMLWithGitHubAlertBlocks(NSString *html)
+{
+    if (!html.length)
+        return html;
+
+    static NSRegularExpression *alertRegex = nil;
+    static NSDictionary<NSString *, NSString *> *titles = nil;
+    static dispatch_once_t alertToken;
+    dispatch_once(&alertToken, ^{
+        NSString *pattern =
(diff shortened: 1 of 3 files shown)


schuyler (changes requested):
Thanks for the PR, @yusufm!

## Issues

- **Nested blockquotes:** an alert containing a nested `> >` quote makes the regex stop at the inner `</blockquote>`, leaving a dangling outer tag and malformed HTML.
- **Single-type test coverage:** only `WARNING` is exercised, so a typo in any of the `NOTE`/`TIP`/`IMPORTANT`/`CAUTION` class or title mappings would go uncaught.
- **Case-insensitivity untested:** the case-insensitive flag (which supports `[!note]`) has no test guarding it against a silent regression.

## Suggestions

- A regression test confirming an alert marker inside a fenced code block stays unconverted would guard against the regex becoming too greedy.
- A multi-paragraph alert body test would document the expected behavior and protect the body-capture logic.
- Gating the transform and its stylesheet behind an `MPRendererDataSource` flag (as other renderer features are) would keep the architecture consistent, though it could also land as a follow-up.
- An `NSAssert` after the regex compile would surface a pattern failure during development rather than silently turning the transform into a no-op.

Issue #397, opened 2026-04-21
Settings - Compilation Tab - Theme and CSS are overlapping and 2 lines translations overflows in. French version


rcuisnier:
On Settings window, Compilation Tab : Theme and CSS form fields are overlapping each others (despite I try to change the window size):

<img width="735" height="581" alt="Image" src="https://github.com/user-attachments/assets/af14af70-0388-4851-83c1-1a591d8c6654" />

I also have it due to French translations taking 2 lines in settings / General Tab

For Automatic Files link creation :
<img width="393" height="427" alt="Image" src="https://github.com/user-attachments/assets/66a01563-d3eb-4ae6-95ee-632f088db8a8" />

And in settings / Editor, last checkbox of behavior form field (the second line is hidden as overflow of the form fieldset container :

<img width="644" height="540" alt="Image" src="https://github.com/user-attachments/assets/f21f533b-53d9-4b3f-bcdf-ff72ba3bcd47" />

boston-black:
I have similiar problems, my settings are all messed up.

15.7.5 (24G624)

Version 3000.0.6 (1360)

<img width="460" height="534" alt="Image" src="https://github.com/user-attachments/assets/fe71e7e0-9ad3-4fdb-8c2b-5b540ea9b824" />

Thanks for this nice tool!

schuyler:
## Requirements & Implementation Plan

Working on this issue. Based on analysis of the codebase, here are the three distinct problems and their root causes:

### Problem 1: Compilation/Rendering Tab — Theme & CSS fields overlapping
**File:** `MacDown/Localization/Base.lproj/MPHtmlPreferencesViewController.xib`
- The CSS and Theme rows have fixed frame positions with insufficient spacing in the XIB
- French label text is longer than English, causing horizontal overflow that pushes elements into adjacent fields

### Problem 2: General Tab — French text wrapping and overflowing
**File:** `MacDown/Localization/Base.lproj/MPGeneralPreferencesViewController.xib`
- Checkbox labels have fixed heights (14–18pt) that don't accommodate 2-line wrapped text
- The "Automatic Files link creation" label in French wraps to 2 lines but the container doesn't expand

### Problem 3: Editor Tab — Last checkbox clipped/hidden
**File:** `MacDown/Localization/Base.lproj/MPEditorPreferencesViewController.xib`
- The behavior fieldset has a hard-coded height constraint of 175px
[...]

PR #419, opened 2026-05-03
Fix preferences layout clipping

yusufm:
Related to #397

## Summary
- Widen the HTML preferences pane so the theme reload/reveal controls no longer clip
- Increase editor preference text field and stepper frames to match current intrinsic sizes
- Verify HTML, Editor, and General preference XIBs have no ibtool notices/warnings/errors

## Tests
- ibtool --warnings --errors --notices MacDown/Localization/Base.lproj/MPHtmlPreferencesViewController.xib
- ibtool --warnings --errors --notices MacDown/Localization/Base.lproj/MPEditorPreferencesViewController.xib
- ibtool --warnings --errors --notices MacDown/Localization/Base.lproj/MPGeneralPreferencesViewController.xib

### DIFF
--- a/MacDown/Localization/Base.lproj/MPEditorPreferencesViewController.xib
+++ b/MacDown/Localization/Base.lproj/MPEditorPreferencesViewController.xib
@@ -23,7 +23,7 @@
             <autoresizingMask key="autoresizingMask" flexibleMaxX="YES" flexibleMinY="YES"/>
             <subviews>
                 <textField verticalHuggingPriority="750" allowsCharacterPickerTouchBarItem="YES" translatesAutoresizingMaskIntoConstraints="NO" id="g0N-qr-H8K">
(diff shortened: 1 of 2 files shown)


schuyler:
Thanks for this, @yusufm — the horizontal clipping you fixed here is real and still relevant (more so now that the long-named "GitHub Dark Default" theme has landed). Unfortunately we created a conflict for you: we merged #461 (localized text wrapping) on top of this, and the two touch the same controls in the HTML preferences XIB.

Rather than ask you to untangle our merge, I've rebased your HTML-pane fix forward onto current `main` in #467 and credited you as co-author. It preserves #461's checkbox word-wrap and adds a regression test guarding the pane width. I left out the editor-pane rect nudges since those were design-time-only values that are inert under Auto Layout and overlapped #461's rework of that pane.

Closing in favor of #467 — but please take a look there and let me know if I missed anything. Much appreciated.

---

Issue #44, opened 2025-11-18
Add configurable column width/wrapping option


schuyler:
## Feature Request
Add the ability to configure column wrapping at a specific character width (e.g., 80 characters).

## Use Cases
- Following team style guides (80 or 120 character limits)
- Maintaining consistent line lengths
- Better version control diffs
- Standard development practices

## Implementation
- Add preference setting for target column width
- Visual indicator (ruler/guide line) at the specified column
- Optional: soft wrap at column width
- Common values: 80, 100, 120 characters

## Source
Based on issue #1337 from MacDownApp/macdown
https://github.com/MacDownApp/macdown/issues/1337

## Priority
MEDIUM - Useful for teams with style guides and development standards

## Implementation Notes
- Add preference for column width setting
- Draw column guide in editor
- Consider soft-wrap option
- Should be optional/configurable

boston-black:
That is a wonderful idea, making Macdown3000 a very competitive tool for focussed writing :) Thanks!

schuyler:
Closing as duplicate of #422, which has a more specific implementation plan (column guide at 80/100/120 with View menu controls and wrap-at-guide option).

PR #422, opened 2026-05-03
Add editor column guide controls

yusufm:
## Summary
- add View menu controls for showing a column guide and wrapping at it
- support 80, 100, and 120 column guide widths
- persist guide/wrap preferences and draw the guide in the editor

## Tests
- xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme MacDown -only-testing:MacDownTests/MPPreferencesTests -only-testing:MacDownTests/MPEditorViewSubstitutionTests -destination 'platform=macOS'\n\nRelated to #44


schuyler (changes requested):
Thanks for the PR, @yusufm!

The shape here is solid — prefs, a guide drawn in `drawViewBackgroundInRect:`, optional hard-wrap, and a View › Column Guide submenu. The wrap path is where most of the risk is.

## Issues

- **Monospace assumption.** Both the guide position (`columnGuideXPosition`) and the wrap width derive from the advance width of `"0"` × columns. That only aligns for a fixed-pitch font; MacDown lets users pick any editor font, so with a proportional font the guide line and wrap point drift from the real character column.
- **Guide line vs. wrap point diverge.** `columnGuideXPosition` includes `textContainerOrigin.x` (the inset), but the wrap container width in `applyColumnGuidePreferences` is `charWidth*column + 2*padding` with no inset term, so when wrapping is on the drawn line and the actual wrap boundary won't coincide.
- **Interaction with existing width-limiting.** `applyColumnGuidePreferences` hard-sets `widthTracksTextView` and `containerSize` in both branches. MacDown already manages editor width via `editorWidthLimited`/`editorMaximumWidth` + `adjustEditorInsets`; forcing the container to `{CGFLOAT_MAX, CGFLOAT_MAX}` when wrap is off can stomp that. Worth confirming width-limited + wrap-at-guide compose.
- **Horizontal scrolling/clipping.** With `widthTracksTextView = NO` and a fixed container wider than the viewport, the text view/scroll view needs to be horizontally resizable or text will clip. Is that configured?

## Suggestions

- The new menu titles (Show Column Guide, Wrap at Column Guide, 80/100/120 Columns) aren't in `Localizable.strings`.
- Test gap: the tests cover `columnGuideXPosition` and the clamp, but not the wrap-mode container sizing — the riskiest part is currently unexercised.
- `applyColumnGuidePreferences` runs twice within one `setupEditor:` pass (inset branch + end of setup); collapsing to once would avoid the double work.
- Only 80/100/120 are reachable from the menu, but `editorColumnGuideWidth` accepts any integer — consider whether to constrain it.

---

Issue #371, opened 2026-04-13
MacDown 3000 Spins Out of Control on Remote File Systems


gurple:
Hello y'all. Let me start by saying that I'm delighted to have recently found this editor. On to the 'issue'.

It seems that MacDown 3000 starts spinning out of control when it's editing (or trying to save when using the in-built macOS file system revision control mechanisms) files on remotely mounted file systems such as SMB, NFS, etc. This is most observed when then FS is a FUSE-mounted SSHFS volume.

Trying to save the file seems to trigger overwhelming FS activities of some sort. The mount collapses and is left stale. There's no FS there at all, in fact. The system below vomits out "Device not configured" warnings. MacDown pops up a notification that, "The document “Foo_Bar.md” could not be saved."

Trying to now close the window (with now unsuccessfully saved changes) prompts warnings that the open file, "is on a volume that does not support permanent version storage. Your changes have been saved and you will not be able to access older versions of this document once you close it. To undo your changes, click Revert." Despite this notice, there is no FS there now. Nothing has been saved as best I can tell.

In this dialog notification I'm given three options:
1) An ambiguous 'Ok'
2) Revert (to what?)
3) Cancel

No matter what option I chose I don't get closer to a state of recovery. If I choose the ambiguous 'Ok' I'm presented with a new dialog stating, "Do you want to save the changes made to the document “Foo_Bar.md”? You can revert to undo the changes since you last saved the document."

Here I'm given three options:
1) Save
2) Revert Changes
3) Cancel

Choosing 'Save' sends me to the FS malfunction where writes can't occur because of the stale mount and device not being configured.
Choosing 'Revert Changes' gives me a dialog that, "The file “Foo_Bar.md” couldn’t be opened. This is the same as if I chose revert initially.
Choosing  'Cancel' brings me back to the open window with unsaved changes.

Normally I'd try to escape this by doing a 'Save as' and putting that in essentially /dev/null. I don't have that option here. I can try 'Duplicate' but that doesn't handover the original open file to the new duplicate. My only apparent escape is to Force Quit.

gurple:
It seems that I can use 'revert' if I go clean up the stale mount and re-mount. This reveals that there are revision files on the volume such as "Foo_Bar.md.sb-3e14f0f7-jsReDQ".

schuyler:
Howdy @gurple and thank you for the report. This is definitely pretty weird. I'll have a look soon-ish at when I have a few minutes.

schuyler:
[...]

PR #424, opened 2026-05-03
Skip file watchers on remote volumes

yusufm:
## Summary
- skip vnode file watching when a path is not on a local volume
- avoid creating document/resource watchers for remote/network files
- cover local and nil watchability checks

## Tests
- xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme MacDown -only-testing:MacDownTests/MPFileWatcherTests -only-testing:MacDownTests/MPResourceWatcherSetTests -destination 'platform=macOS'

Related to #371

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -3228,6 +3228,9 @@ to link outside that scope.", \
 
     [self stopFileWatching];
 
+    if (![MPFileWatcher canWatchPath:self.fileURL.path])
+        return;
+
     __weak MPDocument *weakSelf = self;
     self.fileWatcher = [[MPFileWatcher alloc]
         initWithPath:self.fileURL.path
--- a/MacDown/Code/Utility/MPFileWatcher.h
+++ b/MacDown/Code/Utility/MPFileWatcher.h
@@ -16,6 +16,9 @@
 /// YES if currently watching.
 @property (nonatomic, readonly, getter=isWatching) BOOL watching;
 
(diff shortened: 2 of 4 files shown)


schuyler (changes requested):
Thanks for the PR, @yusufm!

## Issues

- **Resource watchers unguarded:** `MPResourceWatcherSet -addWatcherForPath:` is a second `MPFileWatcher` instantiation site with no `canWatchPath:` check, so resource files (images/CSS/scripts) on remote volumes still hit the same failure this PR fixes for documents.
- **Non-existent path regression:** `NSURLVolumeIsLocalKey` requires the file to exist, so `canWatchPath:` returns NO for a path that transiently doesn't exist (e.g. after an external editor's atomic rename-replace save), which can leave auto-reload disabled — falling back to the parent directory for the locality check would avoid this.
- **Stale watcher on early-return:** the new early-return in `MPDocument -startFileWatching` runs before `stopFileWatching`, so a previously-armed watcher can leak if `canWatchPath:` returns NO for an already-watched session.

## Suggestions

- **Branch coverage:** the empty-string and non-existent-path branches of `canWatchPath:` are untested and quick to cover in CI.
- **Diagnostics:** passing an `NSError *` instead of `NULL` to `getResourceValue:forKey:error:` and logging it would make the "Device not configured" case from #371 easier to diagnose.
- **Doc comment:** the header comment describes only the local-volume case, though the method also returns NO for nil, empty, and non-existent paths.


---

Issue #408, opened 2026-04-29
Can multi select and open with MacDown 3000 opens too many windows


sandundev:
while this not a bug, it will nice that editor can open one window and let user view other .MD files similar to the Mac Preview app for images/docs.

schuyler:
I believe that functionality already exists. You can open multiple
documents in separate windows. Or did you have something else in mind?

sandundev:
<img width="2572" height="1138" alt="Image" src="https://github.com/user-attachments/assets/f582b13b-350c-4706-b14e-e3a40277480a" />

i meant something like this. on the finder window, user would select multiple .MD files and select to open with MacDown 3000. then instead of opening files in multiple windows, open a window and left panel has the list of file names to select similar to a IDE. you could also have tabs like in sublime Text.
If user opened only a single file then you can have your current default view without that left panel.

PR #425, opened 2026-05-03
Open launch files in window tabs

yusufm:
## Summary
- open pending launch files sequentially so their windows can be grouped
- add additional files from a multi-file open as native macOS tabs
- keep single-file launch opens on the normal document path

## Tests
- xcodebuild build -workspace "MacDown 3000.xcworkspace" -scheme MacDown -destination 'platform=macOS'

Related to #408

### DIFF
--- a/MacDown/Code/Application/MPMainController.m
+++ b/MacDown/Code/Application/MPMainController.m
@@ -25,6 +25,19 @@
 
 static NSString * const kMPTreatLastSeenStampKey = @"treatLastSeenStamp";
 
+typedef void (^MPOpenPendingFileCompletion)(NSDocument *document);
+
+
+NS_INLINE NSWindow *MPWindowForDocument(NSDocument *document)
+{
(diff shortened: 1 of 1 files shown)


schuyler (changes requested):
Thanks for the PR, @yusufm!

## Issues

- **Nil document on open failure:** when a document fails to open, the nil result flows into `MPWindowForDocument` and silently breaks tab grouping for every later file in the batch.


## Suggestions

- **Global tabbing flag:** `setAllowsAutomaticWindowTabbing:YES` runs unconditionally on every activation and persists for the whole session, so scoping it to the multi-file case would avoid overriding the user's system setting.
- **Pending list cleared early:** `filesToOpen` is cleared before any open completes, so a failure or crash mid-open would lose the remaining files.
- **Brief comments:** a sentence on the recursive chain and the "nil `tabbedToWindow` means this file becomes the anchor" invariant would aid future readers.

Issue #429, opened 2026-05-14
Manual anchor links to headings don't navigate in preview


falcon-enoc:
### TL;DR
Standard Markdown anchor links like `[Go](#section)` don't navigate to their target heading in the preview. Headings are rendered without `id` attributes (except in a fragile sequential `toc_N` mode tied to the "Detect TOC token" preference), breaking GitHub-compatible navigation.

### Quick repro
Drop this into a new MacDown window with "Detect TOC token" **OFF** (Preferences → Rendering):

```markdown
[Go to section](#hello)

## Hello
This should be the navigation target.
```

**Expected**: clicking "Go to section" scrolls the preview to `## Hello`.
**Actual**: nothing happens. The rendered HTML contains `<h2>Hello</h2>` with no `id` attribute.

### Behavior matrix

| "Detect TOC" preference | `[link](#toc_N)` | `[link](#slug)` (GitHub-style) |
|---|---|---|
| ON | works | doesn't work |
| OFF | doesn't work | doesn't work |

### Root cause
In `MacDown/Code/Document/MPRenderer.m` (around lines 419 and 772), MacDown passes `tocLevel = hasTOC ? 6 : 0` to Hoedown's HTML renderer. Hoedown's default `rndr_header` (`Pods/hoedown/src/html.c` lines 215–229) only emits `id="toc_N"` when `nesting_level > 0`. With the preference OFF, no heading receives an id. Even with it ON, the IDs are sequential (`toc_0`, `toc_1`, …) rather than derived from the heading text — incompatible with the GitHub/GFM convention and brittle to heading reordering.

### Environment
- macOS 26.4.1 (build 25E253)
- MacDown 3000 v3000.0.4 (installed via Homebrew)
- Also reproduced in a local build of `main` (commit `75c18c3`)

### Status
I've reproduced this consistently and have a working local fix that:
- emits slug-based IDs (derived from heading text) for **all** headings, independent of the preference
- updates the visible `[TOC]` renderer to use the same slugs in its links (so the existing TOC feature keeps working)
- applies the equivalent patch to the Quick Look renderer in `MacDownCore/MPQuickLookRenderer.m` — discovered during the test suite run that it has a parallel Hoedown setup which was also affected

Full test suite (875 tests) passing locally. PR coming shortly. Open to feedback on the approach.

schuyler:
👋 @falcon-enoc — A fix for this is available for testing in **MacDown 3000.0.7-rc.1**.

[...]

PR #430, opened 2026-05-14
Emit slug-based heading IDs for anchor link navigation

falcon-enoc:
## Summary
- Headings now receive text-derived `id` attributes regardless of the "Detect TOC token" preference, enabling standard `[text](#section)` anchor links in the preview.
- The visible `[TOC]` renderer uses the same slugs.
- Applied to both the main preview (`hoedown_html_patch.c`) and the Quick Look renderer (`MPQuickLookRenderer.m`) for parity.

## Tests
- `xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme MacDown -destination 'platform=macOS'` — 875 / 875 passing locally.
- Four new tests in `MPMarkdownRenderingTests.m` for slug behavior.
- 28 golden fixtures regenerated mechanically (uniform `id="..."` additions only).

Related to #429

### DIFF
--- a/MacDown/Code/Document/MPRenderer.m
+++ b/MacDown/Code/Document/MPRenderer.m
@@ -423,6 +423,7 @@ NS_INLINE hoedown_renderer *MPCreateHTMLRenderer(MPRenderer *renderer, int tocLe
         flags, tocLevel);
     htmlRenderer->blockcode = hoedown_patch_render_blockcode;
     htmlRenderer->listitem = hoedown_patch_render_listitem;
+    htmlRenderer->header = hoedown_patch_render_header;
 
     hoedown_html_renderer_state_extra *extra =
         hoedown_malloc(sizeof(hoedown_html_renderer_state_extra));
--- a/MacDown/Code/Extension/hoedown_html_patch.c
+++ b/MacDown/Code/Extension/hoedown_html_patch.c
@@ -161,6 +161,85 @@ void hoedown_patch_render_listitem(
 	HOEDOWN_BUFPUTSL(ob, "</li>\n");
 }
 
+// Build a stable text-derived slug from a heading's HTML content.
+// Strips HTML tags, lowercases ASCII, converts spaces to hyphens,
+// drops ASCII punctuation, preserves UTF-8 multi-byte sequences so
+// accented characters survive (e.g. "Introducción" -> "introducción").
+static void slugify(hoedown_buffer *out, const hoedown_buffer *content)
+{
+    if (!content || !content->size)
+        return;
+
+    int in_tag = 0;
+    int last_was_dash = 1;
+
+    for (size_t i = 0; i < content->size; i++)
+    {
+        uint8_t c = content->data[i];
+
+        if (in_tag)
+        {
+            if (c == '>') in_tag = 0;
+            continue;
+        }
+        if (c == '<')
+        {
+            in_tag = 1;
+            continue;
+        }
+
+        if (c >= 0x80)
+        {
+            hoedown_buffer_putc(out, c);
+            last_was_dash = 0;
+            continue;
+        }
+
+        if (c >= 'A' && c <= 'Z')
+            c += 32;
+
+        if ((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '_')
+        {
+            hoedown_buffer_putc(out, c);
+            last_was_dash = 0;
+        }
+        else if (c == ' ' || c == '\t' || c == '-')
+        {
+            if (!last_was_dash)
+            {
+                hoedown_buffer_putc(out, '-');
+                last_was_dash = 1;
+            }
+        }
+    }
+
+    while (out->size > 0 && out->data[out->size - 1] == '-')
+        out->size--;
+}
+
+// rndr_header replacement that always emits a text-derived id, independent
+// of the TOC nesting level. Enables [link](#section-name) navigation.
+void hoedown_patch_render_header(
+    hoedown_buffer *ob, const hoedown_buffer *content, int level,
+    const hoedown_renderer_data *data)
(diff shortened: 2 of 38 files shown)


schuyler (changes requested):

## Issues
- **HTML entities in slugs:** because slugs are derived from the rendered HTML content, a heading like `Q&A` becomes `id="qampa"` (the literal `amp` leaks in), so the `#qa` anchor a reader would naturally write silently fails to navigate — the same happens for `<`, `>`, `"`, `'` (e.g. `Tips & Tricks` → `tips-amp-tricks`). Since rendered content always escapes a literal `&` to `&amp;`, skipping `&…;` entities the same way the loop already skips `<…>` tags would resolve it, for example near the top of the `slugify` loop:

  ```c
  if (c == '&') {            // skip an HTML entity like &amp; / &lt; / &#39;
      while (i + 1 < content->size && content->data[i + 1] != ';')
          i++;
      i++;                    // consume the ';'
      continue;
  }
  ```

  (The same change would apply to `mp_quicklook_slugify`.)

## Suggestions
- **Duplicate-heading test:** a test pinning current behavior for identical headings (e.g. `## C` and `## C++` both yielding `id="c"`) would document the slug contract.
- **TOC round-trip test:** since this PR changes both sides, a test asserting a heading's `id` equals the matching `[TOC]` `href` would lock in the navigation guarantee.
- **Non-ASCII anchors:** UTF-8 bytes are emitted into `id`/`href` without percent-encoding, which works in modern browsers but isn't canonical.

---

falcon-enoc:
Thanks for the thorough review, @schuyler! All points addressed.

**HTML entities in slugs (Issue)** — Fixed. The slug loop now skips `&…;` entities the same way it already skipped `<…>` tags, in both `slugify` (`hoedown_html_patch.c`) and `mp_quicklook_slugify` (`MPQuickLookRenderer.m`). Now `## Q&A` → `id="qa"`, `## Tips & Tricks` → `tips-tricks`, `## A < B` → `a-b`.

**Duplicate-heading test (Suggestion 1)** — Added `testDuplicateHeadingsCollapseToSameId` pinning that `## C` and `## C++` both collapse to `id="c"`.

**TOC round-trip test (Suggestion 2)** — Added `testTOCHrefMatchesHeadingId` asserting the `[TOC]` `href` equals the heading `id` for the same text, locking in the navigation guarantee.

**Non-ASCII anchors (Suggestion 3)** — Left UTF-8 as-is on purpose: browsers match the URL fragment against the `id` literally (they don't percent-decode the `id` for matching), so encoding `id="café"` to `id="caf%C3%A9"` would break navigation rather than fix it. Documented the rationale in the slugify doc comments. Happy to open a separate issue to explore canonical encoding if useful.

Also added 3 analogous tests for the Quick Look renderer to keep both code paths in parity.

All tests green: 86 passed, 0 failures.

schuyler (approved):
Thanks for the PR, @falcon-enoc — and for the thorough write-up in #429!

This is a clean, well-targeted fix: emitting text-derived IDs unconditionally is the right call, the TOC href and heading ID staying in sync is a nice touch, and the new tests around HTML-entity handling cover a genuinely tricky path. Approving.

A handful of minor, non-blocking polish ideas came up during review (a few extra edge-case tests, a now-unused TOC counter, and a comment tweak) — none need to hold this up, so we've logged them on our side in #471 rather than expand the scope here. Much appreciated!

schuyler:
Thanks again for this, @falcon-enoc! The one red check here isn't a problem with your code — it's a branch-staleness collision, and the fix is mechanical.

**What's happening:** the only failing test is `MPMarkdownRenderingTests testTables`. This branch is based on `main` at `75c18c3`, and since then #440 (`c34f75f`, "restore natural table width and horizontal scrolling") added a new **"Wide Table (many columns, issue #432 regression)"** section to `MacDownTests/Fixtures/tables.md` and `tables.html`. CI runs against your branch *merged into current `main`*, so the rendered `tables.md` now includes that new heading — and your `header` callback (correctly!) gives it a slug `id`, while the merged `tables.html` fixture still has it as a plain `<h2>`. The exact-match golden comparison then fails. It passes locally for you because your branch predates #440 and never had that section.

**The fix:** merge (or rebase onto) the latest `main` and regenerate the golden fixtures so the new heading picks up its `id`. Concretely, `tables.html` just needs:

```diff
- <h2>Wide Table (many columns, issue #432 regression)</h2>
+ <h2 id="wide-table-many-columns-issue-432-regression">Wide Table (many columns, issue #432 regression)</h2>
```

Regenerating via the suite's `REGENERATE_GOLDEN_FILES` path will produce that automatically, and it's worth a quick re-run afterward in case `main` has added headings to any other fixtures since you branched. No changes to the slug logic itself are needed — that part's working exactly as intended.

(The `Run Unit Tests` jobs also log a `403` from the coverage-comment step — that's just the read-only token on fork PRs and is already marked `continue-on-error`, so you can safely ignore it.)

---

falcon-enoc:
Thanks, @schuyler — your diagnosis was spot on.

I rebased onto the latest `main` (picks up `#440` and everything through `#473`) and regenerated the golden fixtures via the suite's `REGENERATE_GOLDEN_FILES` path. As you predicted, **only `tables.html` needed updating**, with exactly the one-line change:

```diff
- <h2>Wide Table (many columns, issue #432 regression)</h2>
+ <h2 id="wide-table-many-columns-issue-432-regression">Wide Table (many columns, issue #432 regression)</h2>
```

No other fixtures picked up new headings from `main`, so the slug logic itself needed no changes. The full suite is green locally.

Pushed in `9b0c13b`.

schuyler:
@falcon-enoc Thanks for your contribution and especially thanks for your patience with all the back and forth. Branch merged!

PR #434, opened 2026-05-21
Fix and clarify line break example in help.md

mjonss:
The example missed the two spaces, actually leading to a line break, which it was trying to show :)


schuyler:
Thanks for catching this, @mjonss! The example was missing the two trailing spaces it was supposed to be demonstrating, so it didn't actually show a line break. The fix is correct and the added labels make the contrast much clearer. Appreciate it. 🙏

---

Issue #432, opened 2026-05-18
Incorrect table columns visualization compared to old Mac Down v0.7.2


dafi:
Hi,

table column are shown too small, below the Mac Down 3000 rendering and the (good) Mac Down v.0.7.2

On old Mac Down too large tables can be scrolled, on 3000 are always rendered on visible width

## Mac Down 3000

<img width="1970" height="597" alt="Image" src="https://github.com/user-attachments/assets/c997e6a3-9a9b-48da-b1c8-2bd1fee2ea87" />

## Mac Down v0.7.2

<img width="1720" height="596" alt="Image" src="https://github.com/user-attachments/assets/1df865cf-05b3-4547-90ff-3d3e19986fd0" />

schuyler:
👋 @dafi — A fix for this is available for testing in **MacDown 3000.0.7-rc.1**.

If you have a moment, please download it, confirm the issue is resolved, and
reply here — your confirmation is what graduates the fix into the next stable
release (currently planned for 2026-07-10).

📦 Download: https://github.com/schuyler/macdown3000/releases/tag/v3000.0.7-rc.1

> First launch: if macOS says the app is from an unidentified developer,
> right-click the app and choose **Open**, then confirm. (This is a signed,
> notarized pre-release build.)

No rush — if we don't hear back, the fix ships anyway. Thanks for reporting it!

dafi:
It's a pleasure for me to be of assistance and, above all, to let you know that the bug is fixed!!

Thank you so much!

schuyler:
@dafi 🙏

Issue #30, opened 2025-11-18
Fix line breaking in HTML exports


schuyler:
## Problem
Long lines in Markdown do not wrap properly when exporting to HTML. Text continues as a single line without breaking, creating poor readability in the exported output.

## Solution
Add the CSS property `word-break: break-word;` to paragraph tags in the HTML export template to enable proper text wrapping for lengthy content.

## Implementation Approach
This is a minimal, CSS-based solution that affects visual presentation without altering the underlying Markdown parsing or content structure. The fix should be applied to HTML export templates/styling.

## Source
Based on PR #1320 from MacDownApp/macdown by @radicm
https://github.com/MacDownApp/macdown/pull/1320

Note: PR #1319 appears to be a duplicate of this same fix.

## Priority
MEDIUM - HTML export is a key feature, and proper text wrapping is essential for document readability.

## Implementation Notes
- Update HTML export templates with proper word-breaking CSS
- Test with long lines and various content types
- Verify exported HTML displays correctly in different browsers

schuyler:
## Requirements Clarification

Based on discussion and architectural review (Groucho consultation), the following requirements have been confirmed:

### Scope
- **Comprehensive approach**: Apply word-breaking to multiple block-level elements, not just `<p>` tags
- **Target elements**: `p`, `li`, `td`, `th`, `blockquote`, `dd`, and `body` (fallback)

### Architecture
- **Universal stylesheet**: Create new `export.css` file in `MacDown/Resources/Extensions/`
- **Pattern**: Follow the successful pattern from issue #28 (print.css for PDF exports)
- **Loading**: Modify `MPRenderer.m` to include export.css in stylesheet cascade

### CSS Properties
```css
word-break: break-word;
overflow-wrap: break-word;
word-wrap: break-word;  /* Legacy fallback */
```

### Application Context
- **Screen preview**: Yes - for UX consistency (users see what they'll export)
- **HTML exports**: Yes - primary target
- **PDF exports**: No direct impact (handled separately by print.css)

### Testing Approach
- **TDD**: Write failing tests first, then implement
- **CI verification**: Push to GitHub Actions for macOS testing
- **Manual testing**: Verify with all 6 themes

### Reference
- Based on PR #1320 from MacDownApp/macdown
- Follows architectural pattern from issue #28 resolution

PR #438, opened 2026-06-01
Fix table rendering regression in preview panel

samqbush:
## Summary

Fixes the table rendering regression where tables display with aggressively compressed column widths in the preview panel.

## Problem

`export.css` applied `word-break: break-word` on `td, th` elements (added for issue #30), causing table columns to wrap text at every opportunity and compress to unusably narrow widths. This rule was never in the upstream MacDownApp/macdown PR #1320 — it was added during the macdown3000 implementation.

## Fix

- **Removed** the `td, th { word-break: break-word }` block
- **Added** table-level overflow rules: `table { display: block; width: max-content; max-width: 100%; overflow: auto; }`

This follows the proven pattern from GitHub-2020.css (the only theme that previously handled tables correctly). Since `export.css` loads last in the stylesheet cascade, this fix applies to all 11 themes.

## Behavior Change

- Wide tables now scroll horizontally instead of compressing columns
- Table cell text still wraps via the `body { overflow-wrap: break-word }` inheritance (gentler than explicit word-break on cells)

Related to #432

### DIFF
--- a/MacDown/Resources/Extensions/export.css
+++ b/MacDown/Resources/Extensions/export.css
@@ -23,10 +23,15 @@ li {
     overflow-wrap: break-word;
 }
 
-/* Table cells - handle long content in tables */
-td, th {
-    word-break: break-word;
-    overflow-wrap: break-word;
+/* Tables - allow horizontal scrolling for wide tables (Issue #432)
+ * Replaces previous td/th word-break rules that caused column compression.
+ * Pattern from GitHub-2020.css which handles table overflow correctly.
+ */
+table {
+    display: block;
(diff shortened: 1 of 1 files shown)


schuyler:
Thanks for tackling #432 from two angles, @samqbush! We went with #440 (now merged) as the solution — it keeps `overflow-wrap` on table cells so long URLs still wrap, restores natural table width on the default theme, and adds a wide-table regression fixture. Closing this one as superseded by #440. Much appreciated — a follow-up is also in progress to extend the same horizontal-scroll behavior to the remaining themes.

---

PR #442, opened 2026-06-03
Add zoom controls for the preview pane

dpankros:
## Summary
Adds zoom controls for the rendered preview pane:

- View menu items: Actual Size (⌘0), Zoom In (⌘+), Zoom Out (⌘-)
- Toolbar dropdown with preset zoom levels
- Presets: 50%, 75%, 90%, 100%, 110%, 125%, 150%, 200%
- ⌘+/⌘- snap to the next preset, Safari-style
- Full page zoom: text, images, and inline elements all scale together
- Zoom level persists across launches via NSUserDefaults
- Existing users pick up the 100% default through preferences migration v6

Implementation uses the legacy `WebView setPageSizeMultiplier:` API
already exposed in `WebView+WebViewPrivateHeaders.h`.

## Test plan
- [x] Unit tests in `MPPreviewZoomTests.m` cover default level, snap up/down
  from on- and off-preset values, min/max clamping, and Actual Size reset
- [x] Full test suite: 879 tests, 0 failures
- [x] Manual smoke test in-app

### DIFF
--- a/MacDown 3000.xcodeproj/project.pbxproj
+++ b/MacDown 3000.xcodeproj/project.pbxproj
@@ -107,6 +107,7 @@
 		BA2B9EC2E95175092A97B41E /* MPResourceWatcherSet.m in Sources */ = {isa = PBXBuildFile; fileRef = E6D070A6E080254A17B3B197 /* MPResourceWatcherSet.m */; };
 		CCD97578A2886FFE73BD96F7 /* MPMathJaxRenderingTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6811F39B13C649FC86C4CE5B /* MPMathJaxRenderingTests.m */; };
 		CHKBOXTGL0001BUILDFILER /* MPCheckboxToggleTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CHKBOXTGL0001FILEREFID /* MPCheckboxToggleTests.m */; };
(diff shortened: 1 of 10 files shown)


schuyler:
Thank you for this, @dpankros — it's genuinely the strongest of the zoom PRs. 🙏

While reviewing the cluster of zoom PRs together, one thing stood out: the other implementations route preview scaling through `scaleWebview`, whose body is actually disabled (`#if 0`, due to an old selection-offset bug), so they only end up zooming the editor font. **Yours is the only one that truly zooms the rendered preview**, because it uses WebKit's `setPageSizeMultiplier:` and sidesteps that broken path — text, images, and layout all scale together. The persistence + migration plumbing and the toolbar preset dropdown are also exactly the right shape.

We've consolidated the design across all the zoom PRs into #470, which unifies editor **and** preview zoom behind a single ⌘+/⌘−/⌘0 key pair, persisted as its own preference. **Your PR is the basis for the preview half of that feature**, and the decision to go with a persisted single preference is built directly on your approach.

No action needed from you — we'll build on your work as the starting point, layer in the editor-font scaling, and handle the menu-shortcut relocation on our side. Leaving this open for now as a building block while #470 comes together. Really appreciate the contribution!

---

schuyler:
Superseded by #552, which implemented unified document zoom covering both editor and preview panes.

---

Issue #431, opened 2026-05-14
Preview pane blank for externally-originated .md files on macOS 26 (distinct from #382)


maskedspitz:
This is a different failure mode from #382 (CRLF handling for Windows-created files). In my case the affected files have LF-only line endings and byte-identical content to working files, yet the preview pane is blank. Likely related to a sandbox/TCC interaction rather than the parser.

**Environment:** MacDown 3000.0.6 (1360), macOS 26.5 (Tahoe), Apple Silicon.

**Reproduction:**

1. Document created and saved entirely inside MacDown 3000, then closed and re-opened (or opened after an app restart): preview renders correctly.
2. `broken.md` — a document with byte-identical content to the working one (verified with a diff tool: 110 chars, 12 lines, UTF-8 no BOM, LF line endings) that originated outside MacDown 3000 and was touched by another sandboxed app earlier in its life: **preview pane is blank, editor pane works fine.** Persists across app restarts.
3. `cat broken.md > clean.md` produces a file with the same content, the same `com.apple.macl`, and the same `com.apple.provenance` bytes as `broken.md`. The cloned `clean.md` renders correctly.

**Notes:**

- File rename does not change the behavior (so it's not a path / URL-escaping issue).
- `xattr -c` does not help, because `com.apple.macl`, `com.apple.provenance`, and `com.apple.lastuseddate#PS` are kernel-protected and survive the clear.
- The two files differ only in inode and birthtime:

```
   $ stat -f "inode=%i birth=%SB mtime=%Sm" broken.md clean.md
   inode=200515389 birth=Jul  8 23:15:47 2025 mtime=May 14 22:02:21 2026
   inode=200515377 birth=May 14 21:24:38 2026 mtime=May 14 22:02:20 2026
```

   `broken.md`'s birthtime is ~10 months older than its mtime, suggesting the file has a long history of being touched by various apps.

**Hypothesis:**
Looks like a sandbox / TCC issue. `broken.md`'s inode+birthtime still matches an old TCC permission record from a previous sandboxed app, and the preview WebView's `file://` load path mishandles this case — silently producing a blank render — while the editor (main process) reads the file fine. The cat-cloned `clean.md` gets a new inode/birthtime, so no stale TCC match, and rendering works.

Note: sharing the broken file itself wouldn't help, since uploading/downloading it would give it a fresh inode and birthtime on the maintainer's machine and presumably make it render correctly. The reproduction trigger lives in filesystem/TCC state, not the file bytes.

[...]

PR #444, opened 2026-06-05
Fix blank preview for Markdown files with the execute bit set

1waterrj:
## Summary
Fixes the blank-preview bug in #431: a Markdown file with the POSIX execute bit set (`chmod +x`) shows an empty preview pane while the editor works normally.

## Root cause
The CVE-2019-12173 navigation guard in `MPDocument.m` vetoes non-user-initiated `file://` navigations that target executables. The catch: legacy WebView reports the app's **own `loadHTMLString:baseURL:` render** to the policy delegate as a main-frame navigation of type "other," whose URL is the base URL — i.e. **the document file itself**. `MPURLSecurityPolicy isExecutableOrAppBundleAtURL:` flags any regular file with an exec bit, so the guard vetoed the entire preview load. (`log stream` shows the repeated `MacDown: Blocked file:// navigation for security` messages during render.)

This is why `chmod -x` fixes it (as discovered in the issue thread), and why affected files tend to be ones that lived on Windows/exFAT/SMB volumes, which routinely surface with `0777` modes.

## Fix
Recognize the in-flight render load — render in progress (`alreadyRenderingInWeb`), main frame, URL equal to the renderer base URL — and let it through before the content-navigation policy runs. **Rendering a document is not executing it.**

The guard is unchanged for content-initiated navigations:
- Clicking a Markdown link to an executable still raises the security alert (re-verified manually)
- Out-of-scope JS navigations are still blocked
- Worst case for a same-URL JS navigation during the brief render window is raw Markdown bytes displayed as text — never execution

## Verification (macOS 26.5 Tahoe, Apple Silicon)
- [x] `chmod +x` file: preview renders (previously blank)
- [x] Normal file: renders, no regression
- [x] Executable-link guard: still fires
- [x] Full test suite: 871 tests pass

Related to #431

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -1193,6 +1193,30 @@ static void (^MPGetPreviewLoadingCompletionHandler(MPDocument *doc))()
             // Otherwise this is somewhere else on the same page. Jump there.
             break;
         default:
+            // Issue #431: Our own loadHTMLString:baseURL: render arrives
+            // here as a main-frame navigation of type "other" targeting the
+            // base URL (the document file itself). Rendering a document is
+            // not executing it, so the render load must bypass the
+            // content-initiated navigation policy below. Without this, a
+            // Markdown file with the POSIX execute bit set (common for
+            // files that lived on Windows/exFAT/SMB volumes) renders a
+            // permanently blank preview, because the executable check
+            // below vetoes the entire page load.
+            //
+            // A render load is identified by: a render in flight
+            // (alreadyRenderingInWeb), targeting the main frame, with the
+            // exact URL the renderer uses as its base. Worst case for a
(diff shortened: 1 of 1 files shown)


schuyler:
Thanks for digging into this, @1waterrj — this is a real bug and your diagnosis of #431 is spot on. It turns out we landed a fix for the same issue in #454 (commit 90d17dd) just before getting to this. That one takes a slightly different tack: instead of bypassing the CVE-2019-12173 navigation guard during render, it rewrites the preview base URL to a non-executable sentinel in the same directory, so WebKit never sees an executable base resource and the guard stays fully intact. Since the underlying blank-preview problem should now be resolved on `main`, I'm going to close this — but please do confirm the fix works for your setup, and reopen/ping us if you still hit a blank preview. Really appreciate the careful writeup on the security trade-offs.

---

Issue #428, opened 2026-05-09
First Open After Install Launches Three Windows


mundijr:
On two different macbooks, I installed macdown-3000 via homebrew.  When launching the app for the first time, three windows opened.  Subsequent open of the app only opens one window.  This behavior exhibited on both macbooks.

psybers:
I observed this behavior.  It looks entirely intentional to me.  It opens a blank window (as you would expect), as well as the `help.md` which provides an introduction to the tool, and (if I recall correctly) the `contribute.md` file.  You would not want those two opening every time, though, so just on a first launch makes sense.

I would suggest this issue can be closed as "expected behavior".

schuyler:
👋 @mundijr — A fix for this is available for testing in **MacDown 3000.0.7-rc.1**.

If you have a moment, please download it, confirm the issue is resolved, and
reply here — your confirmation is what graduates the fix into the next stable
release (currently planned for 2026-07-10).

📦 Download: https://github.com/schuyler/macdown3000/releases/tag/v3000.0.7-rc.1

> First launch: if macOS says the app is from an unidentified developer,
> right-click the app and choose **Open**, then confirm. (This is a signed,
> notarized pre-release build.)

No rush — if we don't hear back, the fix ships anyway. Thanks for reporting it!

schuyler:
Addressed in v3000.0.7 via #445 ("Open only the help document on first launch").

mundijr:
Apologies, my notifications weren't working. Thank you. Keep up the great work 🙂

PR #445, opened 2026-06-06
Open only the help document on first launch

1waterrj:
## Summary
Addresses #428: first launch opened three windows (untitled + `help.md` + `contribute.md`), which users on multiple machines reported as confusing.

## Change
`showFirstLaunchTips` now opens only `help.md`. With the standard untitled document that makes two windows. Contributing remains one click away in the Help menu (`Help → Contribute`), where returning users — the ones actually in a position to contribute — can find it.

For context: the three-window behavior was inherited from classic MacDown (contribute.md was added to the first-launch tips in 2017), so this is a deliberate UX change, not a regression fix. Happy to adjust if you'd prefer different first-launch behavior.

## Verification (macOS 26.5 Tahoe, Apple Silicon)
- [x] Simulated fresh install (cleared `firstVersionInstalled`): exactly two windows — untitled + help.md
- [x] Second launch: only the untitled document, tips don't re-fire
- [x] Full test suite passes (871 tests)

Related to #428

### DIFF
--- a/MacDown/Code/Application/MPMainController.m
+++ b/MacDown/Code/Application/MPMainController.m
@@ -354,8 +354,11 @@ NS_INLINE void treat()
 
 - (void)showFirstLaunchTips
 {
+    // Issue #428: Open only the help document on first launch. Together
+    // with the standard untitled document this makes two windows; also
+    // opening contribute.md made it three, which users reported as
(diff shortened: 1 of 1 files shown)


schuyler:
@1waterrj thanks very much for tackling this!

Issue #471, opened 2026-06-24
Follow-ups from PR #430: heading slug anchor IDs (polish, edge-case tests, dedup decisions)


schuyler:
Tracking the non-blocking items surfaced while reviewing PR #430 (slug-based heading anchor IDs). None of these block that PR — capturing them here so we can handle them ourselves rather than asking the contributor to expand the scope of a focused fix.

## Test coverage gaps
- [ ] Empty-slug `"section"` fallback — no test exercises the branch where `slugify()` produces nothing (e.g. a punctuation-only `## !!!` or dash-only `## ---` heading).
- [ ] Purely non-ASCII heading (e.g. `## 你好`) — distinct from the mixed `Introducción` case; exercises the id as a raw UTF-8 byte sequence.
- [ ] Repeated-dash collapse and trailing-dash trim (e.g. `## Foo --- Bar` → `foo-bar`, `## Foo ` → `foo`) — each is its own code path with no current coverage.
- [ ] Digit/underscore passthrough (e.g. `## Version 2_0` → `version-2_0`).
- [ ] Harden the entity-skip tests with negative assertions (e.g. assert `id="qampa"` is absent, not just `id="qa"` present).
- [ ] `testDuplicateHeadingsCollapseToSameId`: add `XCTAssertGreaterThan(secondC, firstC)` for rigor.
- [ ] Quick Look anchor tests are a subset (3) of the preview suite (9); add the fallback and pure-UTF-8 parity cases (especially relevant given the duplicated implementation, below).

## Code clarity
- [ ] Remove the now-unused `header_count++` in `hoedown_patch_render_toc_header` — the value is no longer read now that the TOC href is slug-based.
- [ ] Fix the "encoding would break navigation" comment in both slug copies — modern browsers decode percent-encoded fragments before matching ids, so encoding would *not* break navigation; the raw-UTF-8 choice is fine, only the stated reason is inaccurate.
- [ ] Note in a comment that `&#39;` / numeric character references are skipped (dropped), not decoded, to prevent a future "fix" that decodes them.
- [ ] Comment the `>= 0x80` well-formed-UTF-8 passthrough assumption and the trailing-hyphen-trim loop.
- [ ] Reword the `hoedown_patch_render_header` comment: "independent of the TOC nesting level" → "independent of the Detect TOC token preference".

## Structural / product decisions
[...]

PR #512, opened 2026-07-04
Align heading slug algorithm with GitHub's github-slugger

falcon-enoc:
## What

Align the heading-anchor slug algorithm with GitHub's `github-slugger`, so
`[text](#slug)` links copied from GitHub-rendered Markdown keep working in the
MacDown preview, exported HTML, and Quick Look.

## Why

Follow-up to #430 (which introduced text-derived heading ids). Using the
feature on a real-world GitHub-style TOC surfaced a functional gap (#503): the
previous `slugify()` let every byte `>= 0x80` pass through raw and only
lowercased ASCII, so Unicode punctuation survived into ids and the anchors
silently did nothing. It also collapsed hyphen runs, which GitHub does not do.

Concretely, before this change:

| Heading | Before | After (= GitHub) |
|---|---|---|
| `Chapter 2 — Getting Started` | `chapter-2-—-getting-started` ❌ | `chapter-2--getting-started` ✅ |
| `Überblick` | `Überblick` ❌ | `überblick` ✅ |
| `Foo --- Bar` | `foo-bar` ❌ | `foo-----bar` ✅ |
| `Introducción` | `introducción` ✅ | `introducción` ✅ (unchanged) |

## How

In the slug loop, decode UTF-8 codepoints instead of treating high bytes as
opaque:

- drop the Latin-1 punctuation/symbol block (U+00A1–U+00BF), the
  multiplication/division signs (U+00D7, U+00F7), and General Punctuation
  (U+2000–U+206F: em/en dashes, curly quotes, ellipsis, …);
- lowercase and re-encode Latin-1 uppercase letters (U+00C0–U+00DE, excluding
  U+00D7);
- pass every other codepoint through unchanged as raw UTF-8;
- malformed UTF-8 falls back to passing the raw byte through, as before.

On the ASCII side, map each space/tab to exactly one hyphen without collapsing
runs, and trim leading/trailing whitespace from the heading text.

The two byte-for-byte identical copies
(`MacDown/Code/Extension/hoedown_html_patch.c` and
`MacDownCore/MPQuickLookRenderer.m`) are kept in sync; consolidating them into
a shared unit is tracked separately in #471.

## Testing

- New parity tests mirrored across the preview (`MPMarkdownRenderingTests`) and
  Quick Look (`MPQuickLookRendererTests`) suites: em/en dash dropping, inverted
  punctuation, Latin-1 lowercasing, no hyphen-run collapse, ordinary-Unicode
  passthrough, punctuation-only `section` fallback, and a verbatim real-world
  heading.
- Full suite green locally (`109` tests, `0` failures), including the golden
  fixture comparison (no fixtures needed regeneration — their headings are
  plain ASCII).
- Verified navigation empirically in a real browser: clicking each rendered TOC
  anchor scrolls to its heading and updates `location.hash`, including the
  raw-UTF-8 ids (`#índice`, `#qué-es-aws`, `#conexión-on-premisesaws`),
  confirming the no-percent-encoding convention works.

## Known limitations (out of scope, tracked in #471)

- **Non-Latin-1 uppercase** (Cyrillic, Greek, Latin Extended — Turkish/Polish/
  Czech/…) is not lowercased; GitHub does. Symbols outside the covered blocks
  (CJK punctuation, arrows, emoji, …) are not dropped.
- **Duplicate-heading disambiguation** (`-1`/`-2` suffixes): identical headings
  still collapse to the same id (only the first is reachable by anchor). This
  is an open product decision in #471, not addressed here.
- **Code consolidation** of the two duplicated slug copies remains in #471.

Related to #503, #430, #471, #429.

### DIFF
--- a/MacDown/Code/Extension/hoedown_html_patch.c
+++ b/MacDown/Code/Extension/hoedown_html_patch.c
@@ -172,22 +172,85 @@ void hoedown_patch_render_listitem(
 	HOEDOWN_BUFPUTSL(ob, "</li>\n");
 }
 
-// Build a stable text-derived slug from a heading's HTML content.
-// Strips HTML tags and skips HTML entities (&amp; / &lt; / &#39; ...),
-// lowercases ASCII, converts spaces to hyphens, drops ASCII punctuation,
-// preserves UTF-8 multi-byte sequences so accented characters survive
-// (e.g. "Introducción" -> "introducción"). Anchors keep their raw UTF-8
-// bytes (no percent-encoding): browsers match the URL fragment against
-// the id literally, so encoding would break navigation.
+// Decode the UTF-8 codepoint starting at data[i] (i < size). Returns the
+// number of bytes consumed (1-4) and stores the codepoint in *codepoint, or
+// returns 0 if the sequence is malformed / truncated, in which case the
+// caller should pass the raw byte through unchanged.
+static size_t decode_utf8_codepoint(const uint8_t *data, size_t size, size_t i,
+                                     uint32_t *codepoint)
+{
+    uint8_t c = data[i];
+    size_t len;
+    uint32_t cp;
+
+    if ((c & 0x80) == 0x00)      { cp = c;        len = 1; }
+    else if ((c & 0xE0) == 0xC0) { cp = c & 0x1F; len = 2; }
+    else if ((c & 0xF0) == 0xE0) { cp = c & 0x0F; len = 3; }
+    else if ((c & 0xF8) == 0xF0) { cp = c & 0x07; len = 4; }
+    else return 0;
+
+    if (i + len > size)
+        return 0;
+
+    for (size_t k = 1; k < len; k++)
+    {
+        uint8_t cc = data[i + k];
+        if ((cc & 0xC0) != 0x80)
+            return 0;
+        cp = (cp << 6) | (cc & 0x3F);
+    }
+
+    *codepoint = cp;
+    return len;
+}
+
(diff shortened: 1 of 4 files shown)


schuyler (approved):

## Suggestions

- **QuickLook test parity:** `testHeadingAnchorIdMatchesGitHubForRealWorldHeading` is the one new test without a Quick Look mirror, and adding one would help catch any future drift between the two hand-synced implementations.
- **Decoder unit tests:** A few direct tests on `decode_utf8_codepoint` — malformed/truncated UTF-8, valid 3-byte/4-byte passthrough, the U+00D7/U+00DE boundary, and back-to-back dropped codepoints — would lock in the current behavior of the newest and most intricate code in the diff.
- **Leading-hyphen edge case:** A heading like "— Título" appears to slug to "-título" since the trim step only strips whitespace bytes at the very start/end of the buffer; a quick test would clarify whether that leading hyphen is intended and matches GitHub's own behavior.
- **Scope-limitation guardrail:** A negative test on an untouched Cyrillic or Greek heading (asserting today's non-lowercased output) would make the documented scope boundary explicit for future readers.
- **Comment accuracy on the decoder:** The doc comment on `decode_utf8_codepoint` reads as if it fully validates malformed sequences, including overlong and surrogate-range cases, which is slightly stronger than what the code actually checks.
- **Stale comment line:** The "so encoding would break navigation" sentence in both doc comments wasn't updated even though the surrounding comment block was substantially rewritten, and it's understood to be inaccurate since modern browsers decode percent-encoded fragments before matching ids.

---

PR #514, opened 2026-07-04
Add heading-slug parity tests and fix decoder comments (#503 review follow-up)

falcon-enoc:
Follow-up to the merged PR #512, addressing the review suggestions left on #503. **No change to the slug algorithm itself** — each divergence was verified against `github-slugger` before being pinned in a test.

## Review suggestions addressed

| # | Suggestion | What changed |
|---|---|---|
| 1 | QuickLook test parity | Ported the missing `RealWorldHeading` mirror and mirrored every new test into `MPQuickLookRendererTests` |
| 2 | Decoder unit tests | Added `decode_utf8_codepoint` coverage through the render path (3-byte CJK, U+00D7/U+00DE boundary, back-to-back drops, 4-byte emoji) |
| 3 | Leading-hyphen edge case | Verified against github-slugger (`— Título` → `-título`) and pinned it — the leading hyphen is intentional GitHub parity |
| 4 | Scope-limitation guardrail | Negative tests documenting that Cyrillic/Greek are **not** lowercased (github-slugger lowercases them; we deliberately scope to Latin-1) |
| 5 | Comment accuracy on the decoder | Rewrote `decode_utf8_codepoint`'s doc comment to state what it actually validates and that it does not reject overlong / surrogate-range values |
| 6 | Stale comment line | Dropped the inaccurate "encoding would break navigation" clause in both `slugify` copies |

## Notes

- The malformed/truncated UTF-8 branch of the decoder is unreachable through the `NSString` render entry point (an `NSString` is always valid Unicode), so it is covered by reasoning rather than an integration test. The reachable branches (valid multi-byte + range boundaries) are tested.
- All parity expectations were captured by running `github-slugger` locally on the same inputs.

## Verification

`xcodebuild test` on macOS — 17 new tests pass across both suites (8 preview + 9 Quick Look, the latter with `ENABLE_QUICKLOOK_TESTS=1`).

Related to #503, #512, #471.

### DIFF
--- a/MacDown/Code/Extension/hoedown_html_patch.c
+++ b/MacDown/Code/Extension/hoedown_html_patch.c
@@ -173,9 +173,13 @@ void hoedown_patch_render_listitem(
 }
 
 // Decode the UTF-8 codepoint starting at data[i] (i < size). Returns the
-// number of bytes consumed (1-4) and stores the codepoint in *codepoint, or
-// returns 0 if the sequence is malformed / truncated, in which case the
-// caller should pass the raw byte through unchanged.
+// number of bytes consumed (1-4) and stores the codepoint in *codepoint.
+// Returns 0 when the lead byte is not a valid UTF-8 start byte, when the
+// sequence is truncated (fewer bytes remain than the lead byte promises), or
+// when a continuation byte is malformed; the caller then passes the raw byte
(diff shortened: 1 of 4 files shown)


schuyler:
Thank you @falcon-enoc !

Issue #470, opened 2026-06-23
Unify zoom into one feature: ⌘+/⌘−/⌘0 scales both editor and preview


schuyler:
## Summary

We currently have **four open zoom PRs** that overlap on the same menu items and shortcuts but split into two half-features. Neither half is complete on its own, and they collide if merged together. This issue proposes consolidating them into a **single "document zoom" feature**: one zoom level, one key pair (⌘+ / ⌘− / ⌘0), scaling **both the editor and the preview together** — which is exactly what #335 originally asked for.

> **Implementation note:** the synthesis described here is maintainer-led. #395 and #442 are starting points to build on — no further work is expected from their authors, and the Format-shortcut relocation below is part of *our* scope, not theirs.

## Background: what the four PRs actually do

The PRs fall into two camps:

| PR | Author | Model | Scales editor? | Scales preview? | Persisted? |
|----|--------|-------|----------------|-----------------|------------|
| #337 | @ssstonebraker | transient per-document | yes | **no (dead stub)** | no |
| #395 | @schuyler | transient per-document | yes | **no (dead stub)** | no |
| #411 | @yusufm | transient per-document | yes | **no (dead stub)** | no |
| #442 | @dpankros | preview-only | **no** | yes (works) | yes |

## Key finding: the "both panes" PRs don't actually zoom the preview

The three per-document PRs (#337, #395, #411) route preview scaling through `MPDocument`'s existing `scaleWebview`, whose body is dead code:

```objc
CGFloat scale = [self previewScale];   // computed…
#if 0
    // Sadly, this doesn't work correctly.
    // It looks fine, but selections are offset relative to the mouse cursor.
    ...                                 // …the code that applies `scale` is #if 0'd out
#endif
```

That `#if 0` predates all of these PRs — MacDown disabled view-transform preview scaling long ago because of a selection-offset bug. So those three PRs compute a preview scale factor and throw it away: **in practice they only zoom the editor font.**

#442 is the only PR that actually zooms the preview, because it sidesteps that broken path entirely and uses WebKit's `setPageSizeMultiplier:` (true page zoom — text, images, and layout), which doesn't have the selection-offset problem.

So the real split is:

- **#337 / #395 / #411** → working **editor-font** zoom; preview zoom is a non-functional stub.
- **#442** → working **preview** zoom; doesn't touch the editor.

[...]

PR #395, opened 2026-04-21
Address issue #335: Add zoom keyboard shortcuts (Cmd+Shift+=/Cmd+Shift+-)

schuyler:
## Summary

Adds zoom keyboard shortcuts for both the editor and preview panes, building on the work Steve Stonebraker (@ssstonebraker) started in #337.

- **Zoom In**: `Cmd+Shift+=`
- **Zoom Out**: `Cmd+Shift+-`
- **Actual Size**: menu-only (no shortcut)

Steve's two original commits are cherry-picked with attribution preserved. On top of that, this PR fixes three bugs uncovered during review and addresses a performance regression found in round-2 review.

## Credit

Commits `5b79801` and `907530c` are Steve Stonebraker's original work, cherry-picked from #337. Thanks for the foundation!

## Bugs fixed on top of the cherry-picked work

1. **Shortcut conflict with existing menu items.** The original PR bound `Cmd+0` to Zoom In (which conflicts with Paragraph), and later attempts with `Cmd+=` / `Cmd+-` conflicted with Format > Highlight and Format > Strikethrough. Settled on `Cmd+Shift+=` / `Cmd+Shift+-`, which are free on macOS and match the zoom shortcuts many other editors use.
2. **Editor zoom reset on font-preference change.** The KVO observer for `editorBaseFontInfo` called `setupEditor:` which read `preferences.editorBaseFont` directly and ignored `zoomMultiplier`, snapping the editor back to 100% whenever the user changed editor font settings. Fixed by introducing a `zoomedEditorFont` helper used in both the setup path and the zoom path.
3. **Tab stops ignored zoom.** Tab stop positions were computed from the unscaled base font width, leaving tabs visually misaligned after zooming. Tab stops are now recomputed from the zoomed font.

## Performance fix

First pass routed `applyCurrentZoom` through `setupEditor:`, which triggered a full highlighter re-parse, stylesheet reload and CALayer rebuild on every zoom action. Extracted a lightweight `applyEditorFontAndParagraphStyle` helper that only touches the font and paragraph style — zoom is now snappy.

## Tests

Added `MPZoomTests.m` with 21 tests covering:
- Zoom multiplier bounds and increments
- Menu validation for the three zoom actions
- KVO observer behavior across `editorBaseFontInfo` changes
- Tab stop computation at various zoom levels

## Test plan

- [x] All unit tests pass (`MPZoomTests` + full test suite) on CI
[...]

PR #518, opened 2026-07-10
Unify editor and preview zoom

leevi2010-cursor:
## Summary

Unifies editor and preview zoom into a single document-level control, as proposed in issue #470.

- Adds Command-Plus, Command-Minus, and Command-0 menu actions for both panes
- Adds a toolbar zoom menu with presets from 50% through 300%
- Persists one shared zoom level and applies it to every open document window
- Preserves editor font traits, tab-stop scaling, word wrapping, and the existing relative preview sizing preference
- Reapplies preview zoom after WebView reloads

This builds on the implementations and retained commit history from #442 and #395.

## Testing

- 93 focused unit tests pass (`MPZoomTests`, `MPPreviewZoomTests`, and `MPToolbarControllerTests`)
- Debug application build succeeds on macOS with Xcode 26.6
- Interface Builder validation passes for `MainMenu.xib`

Related to #470 and #335.

### DIFF
--- a/MacDown 3000.xcodeproj/project.pbxproj
+++ b/MacDown 3000.xcodeproj/project.pbxproj
@@ -119,6 +119,7 @@
 		BA2B9EC2E95175092A97B41E /* MPResourceWatcherSet.m in Sources */ = {isa = PBXBuildFile; fileRef = E6D070A6E080254A17B3B197 /* MPResourceWatcherSet.m */; };
 		CCD97578A2886FFE73BD96F7 /* MPMathJaxRenderingTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 6811F39B13C649FC86C4CE5B /* MPMathJaxRenderingTests.m */; };
 		CHKBOXTGL0001BUILDFILER /* MPCheckboxToggleTests.m in Sources */ = {isa = PBXBuildFile; fileRef = CHKBOXTGL0001FILEREFID /* MPCheckboxToggleTests.m */; };
+		PRVZMTST0001BUILDFILER /* MPPreviewZoomTests.m in Sources */ = {isa = PBXBuildFile; fileRef = PRVZMTST0001FILEREFID /* MPPreviewZoomTests.m */; };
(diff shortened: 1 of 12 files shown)


schuyler (commented):

## Issues

- **Toolbar zoom dropdown does nothing:** the popup menu items are assigned `target = self.document` inside `setupToolbarItems` (called from `-init`), where the `document` outlet is still nil, so clicking a preset is a silent no-op — the deferred-dispatch proxy pattern already used elsewhere in this file avoids exactly this.
- **Format shortcuts lost, not relocated:** MainMenu.xib strips the key equivalents from Highlight (⌘=), Strikethrough (⌘−), and Paragraph (⌘0) rather than reassigning them per #470, so those commands lose their shortcuts entirely.
- **Two contradictory zoom test suites:** MPZoomTests.m asserts flat ±0.1 stepping on a per-document `zoomMultiplier`, while MPPreviewZoomTests.m asserts preset-snapping on a shared preference — only the latter matches the implementation, and the stale suite would benefit from reconciliation or removal.
- **Headline behavior untested:** nothing verifies that zoom is actually shared across open document windows (the PR's core claim), and such a test would also have surfaced the nil-target bug above.

## Suggestions

- `resetZoom:`/`selectDocumentZoom:` write `documentZoomLevel` directly, bypassing the clamp in `setZoomMultiplier:`, so routing through the setter would keep bounds enforcement in one place.
- The zoom preset array is duplicated between MPToolbarController.m and MPDocument.m with a hand-sync comment, and a single shared source would avoid drift.
- The new `toggleEditorPane:` → `editorStartInPreviewMode` change appears to be PR #519's feature bundled in here, so it's worth confirming that's intentional or splitting it out.
- A dealloc round-trip test for the toolbar controller's new KVO observer would guard the add/remove lifecycle.
- help.md has no zoom section and doesn't note the moved Format shortcuts, leaving users no in-app reference for either.
- Several tests early-`return` when headless, silently passing with zero assertions, where XCTSkip would make those skips visible in CI.

schuyler:
Superseded by #552, which rebased this branch onto current `main` and added fixes for the toolbar wiring, keybinding collisions, and test coverage.

---

PR #519, opened 2026-07-10
Remember editor pane visibility

leevi2010-cursor:
## Summary

Remember the user's editor pane choice when they use **View > Hide Editor Pane** or **Restore Editor Pane**.

- Hiding the editor enables the existing "start in preview mode" preference
- Restoring the editor disables that preference again
- Only updates the preference when the pane visibility actually changes
- Leaves preview-pane toggles and the existing last-visible-pane protection unchanged

This makes the next document window honor the most recent explicit editor-pane choice without adding another preference.

## Testing

- `MPPaneToggleTests`: 32 tests pass
- Adds a non-headless layout test for hide, remember, restore, and clear

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -2450,7 +2450,10 @@ static BOOL MPScanFenceMarker(NSString *line, unichar *outChar, NSUInteger *outL
 
 - (IBAction)toggleEditorPane:(id)sender
 {
+    BOOL wasVisible = self.editorVisible;
     [self toggleSplitterCollapsingEditorPane:YES];
+    if (self.editorVisible != wasVisible)
+        self.preferences.editorStartInPreviewMode = !self.editorVisible;
 }
 
 - (IBAction)toggleAutoSave:(id)sender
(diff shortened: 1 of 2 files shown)


schuyler (changes requested):

## Issues

- **Overloads a user-facing preference:** hiding or restoring the editor pane silently writes `editorStartInPreviewMode` — the explicit "Start in preview mode" Preferences checkbox — so a user's deliberate opt-in gets cleared (and the checkbox visibly flips) as a side effect of a View-menu action, where a dedicated key or window state restoration would keep the two concepts separate.
- **Untested no-op interaction:** no test covers the case where the last-visible-pane guard blocks the toggle, which is the interaction most likely to silently corrupt the remembered preference.

## Suggestions

- A one-line comment at the call site explaining the "write only on actual change" gate and its relationship to the last-visible-pane guard would make the coupling legible.
- The asymmetry between `toggleEditorPane:` (writes the preference) and `togglePreviewPane:` (doesn't) is unexplained, so a note on the intent would help.
- A test asserting `togglePreviewPane:` never touches `editorStartInPreviewMode` would lock in that intentional asymmetry.

PR #532, opened 2026-07-18
Address issue #503: deduplicate duplicate heading slugs (github-slugger semantics)

falcon-enoc:
## Summary

Implements Decision 1 from the issue: duplicate heading slugs are now deduplicated with github-slugger semantics (`-1`, `-2`, … suffixes), so repeated heading text no longer produces colliding anchor `id`s and `[TOC]` `href`s that silently resolve to the wrong heading.

- The dedup is added to the single shared slug emitter `put_heading_slug` in `Dependency/cmark-gfm/src/mdmark.c`, so it applies identically to the HTML heading-`id` pass and the `[TOC]` `href` pass.
- Each pass carries its own per-render occurrences map — a small `cmark_mem`-allocated linear `(slug, count)` list — threaded as a stack-local. This gives per-render reset for free and adds no new exported symbols (`mdmark.h` is unchanged).
- The registration follows github-slugger's `while`-loop exactly (keyed on slug existence, base counter increments, generated slug seeded with count 0), so it also handles the case where a heading's literal text equals an already-generated suffixed slug (`Hello` / `Hello 1` / `Hello` → `hello`, `hello-1`, `hello-2`).
- The base slug algorithm `mdmark_slugify` is untouched — Unicode lowercasing stays Latin-1-scoped per Decision 2 (#471).
- The `[TOC]` pass now advances the dedup counter for **every** heading in document order **before** the `level > toc_level` filter (which still gates only list-item emission), so a heading deeper than the TOC level that shares a base slug cannot desync TOC `href`s from the `id`s the HTML pass emits.

## Related Issue

Related to #503. Interpretation posted before implementation: https://github.com/schuyler/macdown3000/issues/503#issuecomment-5009707740

## Judgment Calls

1. **PR base is `dev/3000.1`, not `main`.** `mdmark.c` exists only on `dev/3000.1` (where the cmark-gfm migration, PR #515, landed at `29b67b0`); `main` is still hoedown-based. This change is branched off and targets `dev/3000.1` accordingly. This is the single place the described code exists, matching the issue's "this lands in `mdmark.c`".
2. **Full github-slugger `while`-loop, not a naive per-base `count++`.** Alignment with github-slugger is the stated goal, and the `while`-loop is the only variant that stays correct when a literal heading collides with a previously-generated suffixed slug.
3. **Dedup state is a stack-local struct threaded through `put_heading_slug`, rather than the file-static `g_checkbox_index` + reset pattern.** The checkbox counter is global only because its value is read across the C↔Objective-C boundary; the dedup map has no external reader, so a stack-local gives per-render reset intrinsically and keeps `mdmark.h` untouched. (This was flagged as an open question in the interpretation; happy to switch to the checkbox pattern if you prefer consistency.)
4. **Quick Look mirror covers id-dedup only, not TOC parity.** `MPQuickLookRenderer` calls only `mdmark_render_html` and has no `[TOC]`/`href` concept, so there is nothing to assert TOC parity against there.
5. **The deep-heading desync guard is currently future-proofing.** In the shipped app `kMPRendererTOCLevel` is hardcoded to 6 and headings never exceed level 6, so the `level > toc_level` branch is unreachable through `MPRenderer`; the counter-before-filter structure is nonetheless the naturally-correct one, and it is exercised by a test that calls `mdmark_render_toc` directly with a smaller `toc_level`.

## Testing

All headless XCTest, run locally (full suite: 1072 executed; the only failures are 2 pre-existing `HGMarkdownHighlighterTests` theme-color assertions unrelated to this change — a local-environment artifact that passes in CI on the same base commit).

New / updated tests:
- `testDuplicateHeadingsDedupIdSequence` — `Hello`/`Hello`/`Hello`/`World`/`Hello` → `hello, hello-1, hello-2, world, hello-3`.
- `testDuplicateHeadingsDedupWhileLoopHandlesSlugCollision` — the literal-vs-suffix collision edge case.
- `testTOCHrefMatchesHeadingIdForDuplicateHeadings` — id/TOC `href` parity across a document with duplicated headings (the fixture requested in the issue).
- `testTOCHrefMatchesIdWhenDeeperDuplicateConsumesSuffix` — direct-C-API guard for the deep-heading case.
- `testEmptyHeadingsDedupWithSectionFallback` — `section`, `section-1`, … for headings that slugify to empty.
- `testDuplicateHeadingDedupIsStableAcrossRenders` — per-render reset / byte-stability.
- `testQuickLookDuplicateHeadingsDedupIdSequence` — Quick Look id-dedup mirror.
- Renamed `testDuplicateHeadingsCollapseToSameId` → `testDuplicateHeadingsGetDedupedIds` (it previously pinned the pre-dedup collision; now asserts `c` / `c-1`).
- Regenerated the `syntax-highlighting-languages` golden fixture, whose `C`/`C++` headings now correctly render `id="c"` / `id="c-1"`.

Note: `Podfile.lock` is intentionally not included — the only local diff there is a `cmark-gfm` SPEC CHECKSUM drift from re-running `pod install` with a different local CocoaPods version. For a `:path` dev pod that checksum hashes the podspec (unchanged here), not the edited source, so CI reproduces the committed value from the unmodified podspec.

### DIFF
--- a/Dependency/cmark-gfm/src/mdmark.c
+++ b/Dependency/cmark-gfm/src/mdmark.c
@@ -41,8 +41,97 @@ static int g_checkbox_index = 0;
 void mdmark_reset_checkbox_index(void) { g_checkbox_index = 0; }
 int mdmark_current_checkbox_index(void) { return g_checkbox_index; }
 
+#pragma mark - Heading slug dedup (issue #503)
+
+// Per-render duplicate-slug dedup, matching github-slugger semantics: a
+// purpose-built linear (slug,count) list, allocated via the render's
+// cmark_mem so it plays nicely with the Linux test harness. Not a global —
+// each render (HTML pass and TOC pass alike) owns its own stack-local
+// instance, so state resets automatically per render/per pass and no
+// mdmark_reset_* export is needed (contrast with g_checkbox_index above,
+// whose value must cross the C<->ObjC boundary and therefore has none of
+// these options).
+typedef struct slug_dedup_entry {
+  struct slug_dedup_entry *next;
+  char  *slug;   // NUL-terminated copy, owned via d->mem
+  size_t len;    // byte length excl NUL
+  int    count;  // occurrences[slug]
+} slug_dedup_entry;
+
+typedef struct slug_dedup {
+  cmark_mem        *mem;
+  slug_dedup_entry *head;
+} slug_dedup;
+
+static void slug_dedup_init(slug_dedup *d, cmark_mem *mem) {
+  d->mem = mem;
+  d->head = NULL;
+}
+
+static slug_dedup_entry *slug_dedup_find(slug_dedup *d, const char *s,
+                                         size_t len) {
+  for (slug_dedup_entry *e = d->head; e; e = e->next)
+    if (e->len == len && memcmp(e->slug, s, len) == 0)
+      return e;
+  return NULL;
+}
+
+static void slug_dedup_put(slug_dedup *d, const char *s, size_t len,
+                           int count) {
+  slug_dedup_entry *e = d->mem->calloc(1, sizeof(*e));
+  e->slug = d->mem->calloc(1, len + 1);
+  memcpy(e->slug, s, len);
+  e->len = len;
+  e->count = count;
+  e->next = d->head;
+  d->head = e;
+}
+
+static void slug_dedup_free(slug_dedup *d) {
+  slug_dedup_entry *e = d->head;
+  while (e) {
+    slug_dedup_entry *n = e->next;
+    d->mem->free(e->slug);
+    d->mem->free(e);
+    e = n;
+  }
+  d->head = NULL;
+}
+
+// Append the deduped form of base[0..base_len) to out, per github-slugger:
+// result = base; while occurrences[result] exists, occurrences[base]++ and
+// result = base + "-" + occurrences[base]; then occurrences[result] = 0.
+// NULL-tolerant: without a dedup map, just pass the base slug through
+// unchanged (used as a defensive fallback; state is never actually NULL on
+// the current call paths).
+static void slug_dedup_emit(slug_dedup *d, cmark_strbuf *out,
+                            const char *base, size_t base_len) {
+  if (!d) {
+    cmark_strbuf_put(out, (const unsigned char *)base, (bufsize_t)base_len);
+    return;
+  }
+
+  cmark_strbuf result = CMARK_BUF_INIT(d->mem);
+  cmark_strbuf_put(&result, (const unsigned char *)base, (bufsize_t)base_len);
+
+  while (slug_dedup_find(d, (const char *)result.ptr, result.size)) {
+    slug_dedup_entry *b = slug_dedup_find(d, base, base_len);
+    int n = ++b->count;
+    char suffix[16];
+    int slen = snprintf(suffix, sizeof(suffix), "-%d", n);
+    cmark_strbuf_clear(&result);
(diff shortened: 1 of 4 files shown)


falcon-enoc:
### Maintainer notes (self-review observations, deliberately not changed)

A self-review surfaced the following minor items. None are correctness/memory/regression issues, so they were left as-is; noting them for your awareness:

- **`state->dedup` deref style (mdmark.c, HTML heading case).** It dereferences `state->dedup` directly, whereas the nearby `CODE_BLOCK` case uses a defensive `state ? ... : ...` guard. Both are effectively dead guards today (`state` is the single, always-non-NULL `renderer->opaque`), so this is only a style inconsistency, not a hardening gap.
- **`slug_dedup_emit`'s `if (!d)` NULL fallback is currently dead code.** Both call sites always pass a live map; the branch is a documented defensive-only fallback. Kept intentionally, flagged so a future reader knows it's unreachable on current paths.
- **Four id-sequence tests use incremental `rangeOfString:` forward-search** (relative-order assertions) rather than the full ordered-array comparison the two parity tests use via `matchesForPattern:`. The current assertions already pin each exact `-N` suffix in order; converting them would be churn on passing tests for a marginal total-count guarantee, so it was deferred.
- **Comment polish** in `testDuplicateHeadingsDedupWhileLoopHandlesSlugCollision` (the explanation of why the third heading becomes `hello-2`) could read more simply; assertions are correct.
- **mdmark.c file-header contract-change list** could gain a one-line pointer to the new dedup machinery to help a future re-diff against a cmark-gfm bump. Cosmetic.

The regenerated `syntax-highlighting-languages` golden fixture (`id="c"` → `id="c-1"` for the `C++` heading) is the expected ripple of the dedup change, cross-checked as correct.

schuyler (approved):

## Suggestions

- **TOC-href test scoping:** in `testTOCHrefMatchesHeadingIdForDuplicateHeadings` the `href="#…"` match spans the whole page and passes only because the fixture has no other anchor, so scoping it to the TOC block (as test #8 does) would harden the PR's core parity check.
- **Exact-array assertions:** tests #3–#6 verify id presence and order but not the absence of unexpected ids, whereas the `matchesForPattern:` + `XCTAssertEqualObjects` style already used in #2/#8 would close that gap.
- **Mirror-image collision case:** a `Hello 1 / Hello / Hello` fixture (the suffix literal registered before any base occurrence) would cover the currently-untested half of the ordering-sensitive dedup logic.
- **Defensive asserts in `slug_dedup_emit`:** asserting on the unreachable `!d` branch and on the `b`-non-NULL invariant before `++b->count` would turn two comment-documented assumptions into fail-fast guards, matching the file's existing `assert()` idiom.
- **Parity contract note:** a line on `mdmark_render_toc`'s declaration that callers pass the same markdown/options as the paired `mdmark_render_html` call would protect the id/href parity from a future desync.
- **`// MacDown:` deviation tag:** the new `put_heading_slug(…, state->dedup)` call site could carry the deviation tag the file banner promises for its contract changes.
- **Local-variable shadowing:** the `slug` strbuf in the TOC loop shadows the `slug` local inside `put_heading_slug`, so a rename such as `heading_slug` would aid readers.
- **"requirements §…" citations:** the `§2.1`–`§4.2b` references in the doc-comment and test comments don't resolve to any document in the repo or the #503 thread, so linking the real decision comment or dropping the numbers would keep the traceability intact.
- **README-MACDOWN.md:** a one-clause mention of the dedup logic in the "MacDown additions" section would help whoever next re-diffs against upstream `html.c`.

---

Issue #541, opened 2026-07-27
Mermaid/Graphviz don't render in HTML export when "include highlighting" is off


schuyler:
## Summary

The HTML-export path has the same coupling that #533 fixed for the live preview: Mermaid and Graphviz diagram scripts are emitted only when highlighting is included, so exporting HTML with **"include highlighting" unchecked** produces a document where diagram fences stay as plain code blocks — no SVG, no error.

This is the follow-up deliberately scoped out of #533 / #540 (which fixed only the live-preview path in `-[MPRenderer scripts]`).

## Root cause

In `-[MPRenderer HTMLForExportWithStyles:highlighting:]` (`MacDown/Code/Document/MPRenderer.m`), the Mermaid/Graphviz script inclusion is nested inside the `if (withHighlighting)` block:

```objc
if (withHighlighting)
{
    // ... prism scripts ...
    // mermaid
    if ([d rendererHasMermaid:self]) { ...mermaidScripts... }
    // graphviz
    if ([d rendererHasGraphviz:self]) { ...graphvizScripts... }
}
```

When `withHighlighting` is `NO`, the mermaid/graphviz scripts are never added to the exported page, so the diagrams never render.

Note this path is reachable independently of the live-preview setting: `withHighlighting` comes from the export panel's "include highlighting" checkbox (`controller.highlightingIncluded` in `-[MPDocument exportHtml:]`), which is only *initialized* from `htmlSyntaxHighlighting` and can be toggled per-export.

## Why it's a separate change from #533

Unlike `-[MPRenderer scripts]`, this method also uses `withHighlighting` to drive `scriptsOption = MPAssetEmbedded` (asset-embedding vs. linking) for a self-contained exported file. So hoisting the mermaid/graphviz blocks out of the gate needs to preserve the correct asset-embedding behavior for those scripts, which is why it wasn't folded into the focused live-preview fix.

## Expected behavior

Exported HTML renders Mermaid and Graphviz diagrams based on their own toggles, regardless of whether "include highlighting" is selected at export time.

## Notes on testing

The existing export-path tests in `MacDownTests/MPRendererEdgeCaseTests.m` (`testRendererWithMermaidEnabled` / `testRendererWithGraphvizEnabled`) only assert the produced HTML is non-nil — they don't check for script presence/absence, so they wouldn't currently catch this. A fix should add assertions that the diagram scripts are present in the exported HTML with `withHighlighting = NO`.

schuyler:
[...]

PR #542, opened 2026-07-27
Address issue #541: render Mermaid/Graphviz in HTML export when highlighting is off

schuyler:
## Summary

Fixes the export-path counterpart of #533/#540: Mermaid and Graphviz diagrams now render in exported HTML regardless of the export panel's "include highlighting" option.

In `-[MPRenderer HTMLForExportWithStyles:highlighting:]`, diagram script inclusion was nested inside the `if (withHighlighting)` gate — which also drives `scriptsOption`. With highlighting off, the diagram scripts were never added **and** `scriptsOption` stayed `MPAssetNone` (so `MPGetHTML` would drop them anyway), leaving exported diagram fences as plain code blocks.

The fix:
- Hoists the Mermaid and Graphviz blocks out of the `withHighlighting` gate into their own top-level `if` blocks.
- Sets `scriptsOption = MPAssetEmbedded` when either diagram type is enabled, mirroring the existing MathJax branch (both parts are required — hoisting alone would leave the scripts silently dropped).
- Leaves Prism gated on `withHighlighting`, and leaves `stylesOption`, the MathJax branch, and the live-preview `-scripts` path untouched.

## Related Issue

Related to #541 (interpretation posted here: https://github.com/schuyler/macdown3000/issues/541#issuecomment-5086226789)

## Judgment Calls

- **Two-part fix, not a pure hoist.** The `scriptsOption` promotion is load-bearing: without it, hoisted scripts render to nothing (`MPAssetNone` → nil → skipped). Mirroring the MathJax branch keeps the idiom consistent.
- **Embedded, not linked.** Export continues to embed diagram JS (`MPAssetEmbedded`), keeping exported files self-contained — consistent with existing export behavior and MathJax.
- **Test assertions use inlined content, never filenames.** Because export embeds (inlines) JS, tests assert on `mermaid.initialize(` and `function doGraphviz(engine) {` (each verified to occur exactly once in the init scripts, absent from the bundled libraries). Filename substrings only work on the linked preview path.
- **Prism asserted absent only.** The Prism submodule/embedded-content substring is version-fragile, so highlighting-off tests assert Prism *absence* (always safe); the highlighting-on regression is guarded by diagram-script no-duplication (occurrence count == 1) rather than a Prism-presence check.
- **Strengthened the two weak tests in place.** `testRendererWithMermaidEnabled`/`testRendererWithGraphvizEnabled` previously asserted only `XCTAssertNotNil` — which is why the bug shipped undetected — so they were upgraded rather than left alongside new tests.

No open questions block this change.

## Testing

Automated (headless, via the mock renderer delegate; the final test gate is the macOS CI run on this branch):
- Strengthened `testRendererWithMermaidEnabled` / `testRendererWithGraphvizEnabled`: diagram scripts embedded and Prism absent with highlighting off.
- New: both diagrams together (highlighting off); each diagram embedded exactly once when highlighting is on (double-inclusion guard); diagrams disabled → no scripts; `withStyles:NO` independence; MathJax coexistence.

---
_Generated by [Claude Code](https://claude.ai/code/session_014HPFyo6fqNTZd8YppidhP7)_

### DIFF
--- a/MacDown/Code/Document/MPRenderer.m
+++ b/MacDown/Code/Document/MPRenderer.m
@@ -890,15 +890,20 @@ NS_INLINE NSString *MPPreviewHeadTags(NSString *checkboxBridgeToken)
         scriptsOption = MPAssetEmbedded;
         [styles addObjectsFromArray:self.prismStylesheets];
         [scripts addObjectsFromArray:self.prismScripts];
-        if ([self.delegate rendererHasMermaid:self])
-        {
-            [scripts addObjectsFromArray:self.mermaidScripts];
-        }
-        if ([self.delegate rendererHasGraphviz:self])
-        {
-            [scripts addObjectsFromArray:self.graphvizScripts];
-        }
-
+    }
+    // Mermaid and Graphviz render from the language-* class (always
+    // emitted) via their own init scripts; they do not depend on Prism.
+    // Keep them out of the highlighting gate so diagrams export even
+    // when syntax highlighting is off (GitHub issue #541).
+    if ([self.delegate rendererHasMermaid:self])
+    {
+        scriptsOption = MPAssetEmbedded;
+        [scripts addObjectsFromArray:self.mermaidScripts];
+    }
+    if ([self.delegate rendererHasGraphviz:self])
+    {
+        scriptsOption = MPAssetEmbedded;
+        [scripts addObjectsFromArray:self.graphvizScripts];
     }
     if ([self.delegate rendererHasMathJax:self])
     {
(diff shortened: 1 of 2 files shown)


schuyler:
## Maintainer notes (from self-review — observed, deliberately not addressed here)

- **Preview/export duplication (follow-up refactor):** the "hoist Mermaid/Graphviz scripts out of the highlighting gate + promote `scriptsOption`" logic now exists in both `-[MPRenderer scripts]` (preview, fixed under #533/#540) and `-[MPRenderer HTMLForExportWithStyles:highlighting:]` (export, this PR), with near-identical comments. That duplication is why the same bug had to be found and fixed twice. A shared private helper (e.g. appending diagram scripts + setting the asset option) would prevent future drift when another diagram type or toggle is added. Left out of this focused fix to keep the diff minimal and low-risk; worth a follow-up.

---

Issue #548, opened 2026-08-02
Bundled stylesheet fixes never reach users who have already launched the app


schuyler:
## Summary

Preview styles are read exclusively from `~/Library/Application Support/MacDown 3000/Styles/`, and the
launch-time copy that populates that directory skips any file already present. The practical result is
that once a user has launched MacDown 3000 even once, **no future fix to a bundled stylesheet will ever
reach them** — silently, with no error and no indication their theme is out of date.

Surfaced while reviewing #547 (adds `img` max-width to four themes, fixing #546). That PR is correct, but
on its own it only helps fresh installs. This is pre-existing and affects all eleven bundled themes
identically — it is not introduced by #547.

## Mechanism

`MPStylePathForName` resolves unconditionally to the Application Support copy, with no bundle fallback:

```objc
// MacDown/Code/Utility/MPUtilities.m:109-117
NSString *MPStylePathForName(NSString *name)
{
    if (!name)
        return nil;
    if ([name hasExtension:kMPStyleFileExtension])
        name = [name stringByAppendingPathExtension:kMPStyleFileExtension];
    NSString *path = MPPathToDataFile(name, kMPStylesDirectoryName);
    return path;
}
```

The bundled files reach that directory only via `copyFiles`, called at launch, which copies a file **only
when the target does not already exist**:

```objc
// MacDown/Code/Application/MPMainController.m:295-306
// Check for existence of each file and copy if it's not there.
for (NSURL *fileSource in contents)
{
    NSString *name = fileSource.lastPathComponent;
    NSURL *fileTarget = [dirTarget URLByAppendingPathComponent:name];
    if ([manager fileExistsAtPath:fileTarget.path])
        [manager copyItemAtURL:fileSource toURL:fileTarget error:NULL];
}
```

So a stale copy is never refreshed and never falls back to the bundle.

For contrast, two nearby loaders do check the user directory first and fall back to the bundle when the
user copy is absent — `MPHighlightingThemeURLForNameInPaths` (`MacDown/Code/Utility/MPUtilities.m:261-285`)
and QuickLook's independent loader (`MacDownCore/MPQuickLookRenderer.m:65-90`). Neither helps here, because
the problem is a copy that *is* present but outdated.

## Reproduction

1. Launch any build of MacDown 3000 once, so `~/Library/Application Support/MacDown 3000/Styles/` is populated.
2. Ship (or check out) a build with a modified bundled stylesheet.
[...]

PR #547, opened 2026-08-01
Add `img` max-width to the four themes that lack it

versuchshaus:
Fixes #546.

Four of the eleven bundled themes — Clearness, Clearness Dark, Solarized (Dark) and Solarized (Light) — had no width constraint on `img`, so images wider than the preview pane rendered at natural size and overflowed horizontally. The other seven already set `max-width` on `img`.

This adds the same constraint to the four:

```css
max-width: 100%;
height: auto;
```

`height: auto` keeps the aspect ratio once the width is clamped.

Placement follows each file's existing structure and indentation rather than appending at the end:

- **Solarized (Dark) / (Light)** — extends the existing `img { margin-bottom: 20px; }` rule (2-space indent).
- **Clearness / Clearness Dark** — a new `img` rule directly after `a img { border: none; }` (4-space indent).

12 added lines, no deletions, no other rules touched.

### Verifying

Before and after, each stylesheet was parsed for a rule whose selector list includes `img` and whose body sets `max-width` (comments stripped). Before: 7 of 11 pass. After: 11 of 11.

To see it by hand: set the style to Solarized (Dark), open a document embedding an image wider than the pane (a 1920x1080 screenshot will do), and compare against a GitHub theme.

Issue #546, opened 2026-08-01
Four bundled themes lack an `img` max-width rule — full-resolution images overflow the preview pane


versuchshaus:
## Summary

Four of the eleven bundled preview themes have no `max-width` on `img`, so any image wider than
the preview pane renders at natural size and runs off the edge. The other seven constrain images
correctly, which makes this look like a document problem rather than a theme problem — the same
file renders fine after switching to a GitHub theme.

Split out from the observation in #491, which currently assesses images as "generally covered"
across themes. That holds for 7 of 11; the four below are the exception.

## Environment

- MacDown 3000 3000.0.7 (1429)
- macOS 26 (Darwin 25.6.0), Apple Silicon

## Reproduction

1. Settings → Rendering → style: **Solarized (Dark)** (or Solarized (Light), Clearness, Clearness Dark).
2. Open a Markdown file embedding an image wider than the preview pane — e.g. a 1920x1080 screenshot: `![shot](shot.jpg)`.
3. The image renders at full natural width and overflows the pane horizontally.
4. Switch to any GitHub theme — the same image is correctly constrained.

## Audit

Parsed each stylesheet in `MacDown/Resources/Styles/` for a rule whose selector list includes
`img` and whose body sets `max-width` (comments stripped):

| Theme | `img` max-width |
|---|---|
| Clearness | **missing** |
| Clearness Dark | **missing** |
| Solarized (Dark) | **missing** |
| Solarized (Light) | **missing** |
| GitHub | present |
| GitHub-2020 | present |
| GitHub Tomorrow | present |
| GitHub2 | present |
| Github2 (dark) | present |
| Gmail | present |
| Google Docs | present |

In `Solarized (Dark|Light).css`, `img` appears only in the reset block
(`margin/padding/font-size/font/vertical-align`) and later as `img { margin-bottom: 20px; }` —
never with a width constraint. `Clearness.css` and `Clearness Dark.css` have no `img` width rule
at all.

## Fix

PR to follow: adds `max-width: 100%; height: auto;` to the four stylesheets, matching what the
GitHub themes already do. `height: auto` preserves the aspect ratio once the width is clamped.

## Workaround for anyone hitting this now

The bundled stylesheets are copied to `~/Library/Application Support/MacDown 3000/Styles/`;
editing the copy there fixes it without touching the app bundle.

Issue #443, opened 2026-06-03
Coloring HTML in the Editor


gregwillits:
More of a tip than an issue...

I was going to ask if there is a way to define the coloring of HTML code in a Theme? Doesn't need to be fancy, everything between <...> can all be one color even. Low priority to parse the tag attributes.

Then I realized there's a style callout for HTML_ENTITY.

So.... HTML ? Yep!

None of the built-in Themes color HTML, but you change that by adding the following in the Theme file. 
```
HTML
foreground: XXXXXX
```
Might be worth updating the default themes.

schuyler:
## Requirements & Approach

Thanks for the tip! You're right — the editor highlighter (peg-markdown-highlight) already recognizes HTML element types that themes can style, but none of the bundled themes do. We're going to fix that.

**Findings**
- The style parser maps a theme section name directly to an element type, so two HTML-related types are available to themes today:
  - `HTML` — inline HTML tags (e.g. `<br>`, `<span>`)
  - `HTMLBLOCK` — block-level HTML (e.g. a `<div>…</div>` block)
- All 15 bundled editor themes define `HTML_ENTITY` (and `COMMENT`) but none define `HTML`/`HTMLBLOCK`, so HTML markup currently renders in the default body text color.

**Decisions**
- **Scope:** Add both `HTML` and `HTMLBLOCK` sections to all 15 bundled themes (not just inline `HTML`), so block-level HTML is colored too.
- **Color choice:** Reuse each theme's existing `HTML_ENTITY` foreground color for `HTML`/`HTMLBLOCK`. HTML tags and entities are sibling constructs, and every theme author already picked a palette-appropriate color, guaranteeing good contrast across light/dark themes.
- **Testing (TDD):** Add a data-driven `XCTest` that enumerates every bundled `.style` theme and asserts each defines an `HTML` and `HTMLBLOCK` style with a non-nil foreground color — a regression guard so future themes don't omit it.

**Out of scope:** No new parser features or attribute parsing (e.g. distinct coloring of tag attributes) — the existing one-color-for-markup behavior the issue describes is sufficient.

Related to #443

---
_Generated by [Claude Code](https://claude.ai/code)_

schuyler:
👋 @gregwillits — A fix for this is available for testing in **MacDown 3000.0.7-rc.1**.

If you have a moment, please download it, confirm the issue is resolved, and
reply here — your confirmation is what graduates the fix into the next stable
release (currently planned for 2026-07-10).

[...]

PR #551, opened 2026-08-02
Address issue #548: refresh bundled styles and themes without destroying user edits

schuyler:
## Summary

Bundled styles and themes are copied into `~/Library/Application Support/MacDown 3000/` only when absent, so every fix we ship to an existing bundled file is invisible to anyone who has already launched the app. #443, #432 and #547 all shipped edits that no existing user has ever received.

The complication is that we invite people to edit these files (`help.md`), and #318 exists specifically so their edits show up live. So the app cannot simply overwrite. It has to distinguish a stale copy it placed itself from one the user wrote — and nothing on disk recorded that.

This adds that record, and refreshes only what it can prove is ours.

- **Provenance manifest** (`BundledResourceManifest.json`, at the data-directory root): the SHA-256 of the bytes the app most recently placed at each path.
- **Shipped-history manifest** (`BundledResourceHistory.json`, in the app bundle): every digest each file has had in a release, generated from the 16 release tags by a checked-in script. This is what rescues files already stale on disk today — without it the fix reaches nobody who currently has the bug.
- A file is refreshed only if its bytes match the provenance entry, the current bundle copy, or a shipped digest. Anything else is the user's and is left alone, with no entry recorded, which is what keeps it protected on later launches.

Scope covers `Styles/` **and** `Themes/` — one loop, one defect, and #443 already bit themes. Prism is unaffected and untouched: it is never seeded here and already falls back to the bundle.

Refreshes copy to a hidden temporary file in the same directory and then replace atomically, so a file is never briefly absent. Entries that are not regular files are skipped without being opened, and the digest helper opens non-blocking and re-checks the descriptor, so a fifo left under a bundled name cannot stall launch. A bundle directory that exists but cannot be read aborts the sync without writing anything, since an unreadable source is indistinguishable from one whose files were all removed and would otherwise drop every entry it covers.

## Related Issue

Related to #548.

The full requirements interpretation, including the decision table this is built and tested against, is on the issue: https://github.com/schuyler/macdown3000/issues/548#issuecomment-5160083149

## Judgment Calls

**Shipping the history manifest, rather than recording provenance only from here on.** Provenance alone cannot classify the files that are already stale, because those are exactly the files with no provenance. Adopting whatever is on disk as provenance is not a smaller version of this fix — it either overwrites real user edits on the second launch, or pins every existing file as unclassifiable forever. So history is load-bearing, and it is why this is larger than the bug first suggests. Its maintenance cost is bounded: it only needs to cover releases up to the one shipping this, and the generator is re-runnable if you ever want to extend it.

**Editor themes included.** The issue names stylesheets; `Themes/` shares the same loop and the same missing fallback, and #443 shipped theme edits nobody received.

**A downgrade reverts pristine files.** Installing an older version restores that version's bundled bytes. Deliberate — pristine files track whatever is installed, and user edits are still protected.

**Never deleting a file we no longer ship.** The user may have it selected as their active style.

**Withdrawn from scope:** a bundle fallback in `MPStylePathForName`/`MPThemePathForName`. It fixes zero stale files and sits on the render path, so it belongs in its own change.

**No "Restore Defaults" button.** Deleting a file and relaunching restores it, which is per-file and already works; a destructive button deserves your judgement on wording and granularity. `help.md` documents the gesture instead.

**Open question for you:** if a user edited a file *and* wants the newer bundled version, they keep their edit and never get the fix. That is the correct side of the trade, but it does mean #546's overflow persists for them until they delete their copy.

## Testing

**Baseline:** 1104 tests, 0 failures on `main` at `7767580` (run 30224808528). **Now:** 1170, 0 failures — 66 added.

- 37 tests drive the sync through injected fake bundle and user roots: all twelve decision-table rows, idempotence, symlinked resource directories, the enumeration guard, permission failures, and manifest key case sensitivity.
- 29 cover the digest helpers against known vectors, both manifest readers, and the classification function's truth table.
- Roots come from a factory that asserts both resolve under the temporary directory before any test runs, so a refactor cannot let a test write to a real Application Support directory.
- Refresh cases assert the target's inode changed and no temporary or backup residue remains, which is what distinguishes an atomic replacement from a truncate in place.
- `HGMarkdownHighlighterTests` reads the real user themes directory during app-hosted runs and is left unmodified deliberately: it is a live canary for the atomic replacement, and it passes.
- The app-launch smoke test exercises the real `copyFiles` path against the runner's home directory and passes on every leg.

One caveat worth stating: the tests were written against the specification before the implementation existed, and were reviewed for it, but the CI run intended to demonstrate them failing never executed — it died on a compile error in the test file instead. That the suite fails without the behaviour rests on review rather than on an observed red run.

`Tools/generate-bundled-resource-history.sh --check` regenerates the history manifest and diffs it against the committed copy, so drift is detectable. Re-run it after cutting a release if you want the newest bytes covered; the four styles touched by #547 are not in history yet because they are not in a tag.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Q79Lih9LxX3EfcJ1LQqN8J)_

### DIFF
--- a/.claude/commands/release.md
+++ b/.claude/commands/release.md
@@ -32,6 +32,7 @@ Use TodoWrite to create a todo list for tracking progress:
 - Validate version and run pre-flight checks
 - Build and commit changelog
 - Create and push git tag
+- Regenerate and commit the bundled resource history
 - Monitor build and notarization
 - Complete release workflow
 ```
@@ -368,6 +369,19 @@ git push origin "v{VERSION}"
 
 If successful, the tag is now on GitHub and the release workflow will start within 1-2 minutes.
 
+#### 3d. Regenerate the Bundled Resource History
+
+The generator enumerates release tags, so this must run *after* the tag is pushed — files shipped in an untagged release can never be classified, which silently removes the recovery path for any user who later loses their provenance manifest.
+
+```bash
+Tools/generate-bundled-resource-history.sh
+git add MacDown/Resources/BundledResourceHistory.json
+git commit -m "Record bundled resource digests for v{VERSION}"
+git push origin main
+```
+
+If the script reports no changes, there is nothing to commit — continue.
+
 
 ### Step 4: Monitor Build Workflow
 
@@ -419,6 +433,7 @@ The website will update automatically after the workflow completes.
 Summary:
 - ✅ CHANGELOG.md and README.md updated and committed
 - ✅ Tag v{VERSION} created and pushed
+- ✅ Bundled resource history regenerated and committed
 - ✅ Build workflow triggered
 
 **What happens next:**
--- a/MacDown 3000.xcodeproj/project.pbxproj
+++ b/MacDown 3000.xcodeproj/project.pbxproj
@@ -160,6 +160,10 @@
 		504AC0000000000000000005 /* MPPDFAnchorInjectorTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 504AC0000000000000000004 /* MPPDFAnchorInjectorTests.m */; };
 		504AC0000000000000000007 /* PDFKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 504AC0000000000000000006 /* PDFKit.framework */; };
 		504AC0000000000000000008 /* PDFKit.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 504AC0000000000000000006 /* PDFKit.framework */; };
+		6DF7634F7CFB0E718623C743 /* MPBundledResourceSync.m in Sources */ = {isa = PBXBuildFile; fileRef = BFB785FC2553E0F066BF4314 /* MPBundledResourceSync.m */; };
+		0EF7F855A6D63BC60A67CCE7 /* BundledResourceHistory.json in Resources */ = {isa = PBXBuildFile; fileRef = BFF7FCCC3D2D303F270D4F89 /* BundledResourceHistory.json */; };
+		7F018F1F3C7619EA26F737AC /* MPBundledResourceSyncTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 93338DE9459DD0BBABF0172B /* MPBundledResourceSyncTests.m */; };
+		276BBE5A1237B16170426444 /* MPBundledResourceDigestTests.m in Sources */ = {isa = PBXBuildFile; fileRef = 90F20FA45DCC9EBE32F401EC /* MPBundledResourceDigestTests.m */; };
 /* End PBXBuildFile section */
 
 /* Begin PBXContainerItemProxy section */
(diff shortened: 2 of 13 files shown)


schuyler:
## Maintainer notes

Things a review of this branch surfaced that were deliberately **not** addressed here, with the reasoning. Nothing below blocks merge on correctness.

### Worth your decision

**Launch cost is now a steady-state cost, and it is unbounded on a network home directory.** The old `copyFiles` did existence checks after first launch — effectively free. This one opens and fully reads up to 56 files (28 bundled, 28 on disk) to compute digests on *every* launch, synchronously in `-[MPMainController init]`, before any window exists. On local storage that is a few milliseconds, which the requirements sized and accepted. On a managed Mac with a redirected network home, a stalled mount hangs launch with no timeout and no file-count ceiling. Nothing can crash — every I/O error is caught and skipped — but the hang is real, and the requirements only ever sized this for local SSD. A cap or a timeout would be new behaviour the requirements do not authorise, so it is your call rather than something to slip in here.

**The generator now has to be re-run at release time.** Both runbooks were updated in this branch, placed after the tag is pushed since the script enumerates tags. But `--check` is still not wired into CI, and the script's own header notes it is cheap enough to be. If a release goes out without regenerating, that release's files are permanently unclassifiable — harmless for most people, but it silently removes the recovery path for anyone who later loses their provenance manifest. Wiring `--check` into `test.yml` would make the omission loud instead of silent. Left alone here because adding a workflow step to a green PR carries its own risk.

**A user who edited a file and wants the newer bundled version still has no automatic route.** They keep their edit and never receive the fix. That is the correct side of the trade, and `help.md` now documents deleting the file and relaunching, but it does mean #546's overflow persists indefinitely for that person. A "restore bundled default" affordance is the obvious follow-up, and was kept out of scope because a destructive action deserves your judgement on wording and granularity.

### Known gaps in the testing

**The red phase was never observed.** The CI run meant to demonstrate the 66 tests failing against stub implementations died on a compile error in the test file instead, and by the time that was fixed the implementation had landed. That the suite fails without the behaviour rests on review, not on a run anyone watched. Several tests carry deliberate companion assertions specifically so a do-nothing implementation cannot satisfy them, but that is an argument, not evidence.

**`MPRefreshFileAtPath`'s failure branch is untested.** It is the only function that overwrites bytes at a path already classified as safe to touch, and no test forces `replaceItemAtURL:` to fail after the copy leg succeeded. The requirements name the immutable flag (`chflags uchg`) as a failure mode to survive; nothing exercises it. A test would need `XCTSkipIf(geteuid() == 0)` and careful teardown or it strands temp directories across four runners, which is why it is not here.

**The generator has no tests at all.** It produces the data that decides whether a file gets overwritten, so a wrong digest there is the one genuine data-loss path in this change. `--check` regenerates and byte-diffs, and the committed manifest was independently re-derived during review, but nothing runs automatically.

### Smaller things

- A concurrent first launch can misreport. Two instances racing to copy the same file mean the loser's `copyItemAtURL:` fails with `NSFileWriteFileExistsError`, which is currently counted as a genuine I/O failure and logged as one. It degrades safely — the next launch backfills — but the accounting and the log line are both wrong. Treating that error as "someone else placed it" would fix it.
- An interrupted refresh can leave a `.<name>.<uuid>.tmp` file that is invisible to every later sync, since hidden files are skipped during enumeration. It requires being killed inside a sub-millisecond window and costs a few KB.
- `MPSHA256HexOfData` is exported from the header with no production caller; only tests use it.
- `MPRefreshFileAtPath` does not clear `*error` on entry, unlike its siblings in the same file.
- `report.aborted` stays `NO` when the data directory itself cannot be created, so that case is indistinguishable from a clean no-op run except through the log.
- The `Styles`/`Themes` pair is spelled independently in the shell script and twice more in its embedded Python, with nothing keeping those in step with the Objective-C constants.
- Files withdrawn from the bundle stay on disk forever by design (the user may have one selected), so over many years and renames the directory only grows.

### Unrelated, noticed in passing

`CLAUDE.md` says the platform is macOS 10.14+. Every `MACOSX_DEPLOYMENT_TARGET` in the project, the `Podfile`, and `README.md` say 11.0.

---

PR #555, opened 2026-08-03
Address issue #548: read bundled preview styles from the app bundle

schuyler:
## Summary

Preview stylesheets were resolved only from `~/Library/Application Support/MacDown 3000/Styles/`, which `copyFiles` populated copy-if-absent at launch. Once a user had launched the app even once, any later fix to a bundled stylesheet never reached them — the stale user copy shadowed the bundle forever, silently (issue #548).

This makes the user Styles directory a pure **override layer** instead of the sole source of truth:

- **Bundle fallback in `MPStylePathForName`** (`MPUtilities.m`): a style not present in Application Support is read straight from the app bundle. All consumers already funnel through this one function, so bundled fixes now reach every user automatically on next launch. Mirrors the existing `MPHighlightingThemeURLForNameInPaths`.
- **Union listing in `loadStylesheets`** (`MPHtmlPreferencesViewController.m`): the stylesheet picker now lists the union of bundle and user styles (user shadows bundle by name), so new bundled themes appear without any copying. Mirrors `MPListHighlightingThemesInPaths`.
- **One-time prune replaces the launch-time copy** (`MPMainController.m` `copyFiles`): instead of copying `Styles/` into Application Support, a user style file is deleted only when its content SHA-256 matches a version that was actually shipped. Those users then transparently fall through to the live bundle; customized files never match and are left untouched. The stock-hash set is a frozen 48-entry list covering every `Styles/*.css` across all release tags on `main` and `3000.1.x`.
- **Reveal-in-Finder guard** (`MPHtmlPreferencesViewController.m`): since the user Styles directory is no longer pre-populated, the "Reveal" action now creates it on demand, matching the existing Prism-theme handler.

The change reuses patterns already in the codebase rather than introducing any resource-sync machinery; net ~200 lines of app code plus unit tests.

## Related Issue

Related to #548. Requirements interpretation posted here: https://github.com/schuyler/macdown3000/issues/548#issuecomment-5169294845

## Judgment Calls

- **Customizations are never touched.** The prune deletes only on an exact content-hash match against a previously shipped version, so any edited file is preserved and keeps shadowing the bundle — honoring `help.md`'s invitation to customize. The failure mode is one-directional: a missed shipped version would simply leave that user's copy in place (today's behavior), never lose data.
- **No new "Reset to defaults" UI, versioning scheme, or update prompt.** With overlay resolution, deleting a customized file already reverts to the current default, so the escape hatch is inherent. The hash set is permanently frozen — after this ships nothing writes stock CSS into Application Support again.
- **Scope is preview styles only.** Editor `Themes/` share the identical latent bug (same copy-if-absent loop) but are deliberately left for a separate change to keep this focused.
- **QuickLook is unchanged** — its independent loader already falls back to the bundle and never ran `copyFiles`.

## Testing

Unit tests added to `MacDownTests/MPUtilityTests.m` (headless-safe — pure functions + filesystem, no window/WebView):

- `MPStylePathForNameInPaths`: user shadows bundle, bundle fallback, `.css` auto-append, missing-from-both fallthrough, nil name.
- `MPListStylesheetsInPaths`: union, dedup, sort, non-css ignored, empty dirs, extension stripping.
- `MPContentHashOfFileAtPath`: known-bytes SHA-256, nil for unreadable.
- `MPPruneStockStylesheetsInDirectory`: stock-match deletion, customized-file survival, mixed-directory count, absent-directory no-op (using test-constructed hash sets, independent of the frozen literal).
- `MPKnownStockStyleHashes`: count is 48 and every entry is 64-char lowercase hex.

The 48-hash set was regenerated and diffed against the source literal during review. Tests run on macOS CI.

---
_Generated by [Claude Code](https://claude.ai/code/session_01CKEPuiEJY4YKKtDmzx2fnN)_

### DIFF
--- a/MacDown/Code/Application/MPMainController.m
+++ b/MacDown/Code/Application/MPMainController.m
@@ -280,8 +280,11 @@ NS_INLINE void treat()
            withIntermediateDirectories:YES attributes:nil error:NULL];
     }
 
+    MPPruneStockStylesheetsInDirectory(MPDataDirectory(kMPStylesDirectoryName),
+                                        MPKnownStockStyleHashesByName());
+
     NSBundle *bundle = [NSBundle mainBundle];
-    for (NSString *key in @[kMPStylesDirectoryName, kMPThemesDirectoryName])
+    for (NSString *key in @[kMPThemesDirectoryName])
     {
         NSURL *dirSource = [bundle URLForResource:key withExtension:@""];
         NSURL *dirTarget = [NSURL fileURLWithPath:MPDataDirectory(key)];
--- a/MacDown/Code/Preferences/MPHtmlPreferencesViewController.m
+++ b/MacDown/Code/Preferences/MPHtmlPreferencesViewController.m
@@ -83,6 +83,14 @@ NS_INLINE NSString *MPPrismDefaultThemeName()
         case 0:     // Reveal
         {
             NSString *dirPath = MPDataDirectory(kMPStylesDirectoryName);
+            NSFileManager *manager = [NSFileManager defaultManager];
+            if (![manager fileExistsAtPath:dirPath])
+            {
+                [manager createDirectoryAtPath:dirPath
+                   withIntermediateDirectories:YES
+                                    attributes:nil
+                                         error:nil];
+            }
             NSURL *url = [NSURL fileURLWithPath:dirPath];
             NSWorkspace *workspace = [NSWorkspace sharedWorkspace];
             [workspace activateFileViewerSelectingURLs:@[url]];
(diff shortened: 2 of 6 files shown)


schuyler:
**Maintainer notes** — things observed during self-review and deliberately left out of this PR, with rationale:

- **Editor `Themes/` have the identical latent bug.** `MPThemePathForName` resolves unconditionally to Application Support and `copyFiles` copies themes copy-if-absent, so a fix to a bundled `.style` never reaches an existing user either. This PR deliberately scopes to preview styles (issue #548); the same overlay+prune pattern would apply. Worth a follow-up issue.
- **`-copyFiles` now also prunes stale styles**, so its name is slightly less precise than before. Left as-is to keep the diff focused; rename if you'd prefer.
- **`loadStylesheets` re-sorts the stylesheet list** with `localizedStandardCompare:` even though `MPListStylesheetsInPaths` already returns sorted output. Harmless and pre-dates this change (the old `MPListEntriesForDirectory` returned unsorted); removing the redundant re-sort would make it consistent with `loadHighlightingThemes` right below it.

None of these affect correctness of the fix.

---

Issue #553, opened 2026-08-03
Re-sign Sparkle's nested components in the release workflow (release-blocking after #129)


schuyler:
## Context

The Sparkle 2 migration for #129 adds the prebuilt `Sparkle.framework` (with nested XPC services and helper tools: Autoupdate, Updater.app, Installer.xpc, Downloader.xpc) to the app bundle via CocoaPods.

The release workflow (`.github/workflows/release.yml`) signs the app with hardened runtime and verifies with `codesign --deep --strict`. Sparkle's nested components ship with Sparkle's own ad-hoc/development signatures and must be **re-signed inside-out with our Developer ID identity and hardened runtime** before the outer app signature is applied, or signing verification and notarization will fail.

**Releases are blocked until this lands.** PR builds and CI test runs are unaffected (Debug, unsigned).

## Required changes

1. In the release workflow's signing step, before signing the outer app, re-sign (inside-out, `--force --options runtime`) each nested Sparkle component:
   - `Sparkle.framework/Versions/B/XPCServices/Installer.xpc`
   - `Sparkle.framework/Versions/B/XPCServices/Downloader.xpc` (if present)
   - `Sparkle.framework/Versions/B/Autoupdate`
   - `Sparkle.framework/Versions/B/Updater.app`
   - then `Sparkle.framework` itself
   (Follow Sparkle's sandboxing/signing documentation for the authoritative component list for the pinned Sparkle version.)
2. Verify `codesign --verify --deep --strict` passes on the assembled app.
3. Confirm notarization succeeds with the re-signed nested components.

## Also needed from a Mac (one-time, can ride along with any Mac-side PR)

- Run `bundle exec pod install` and commit the regenerated `Podfile.lock` plus the CocoaPods-generated `[CP] Embed Pods Frameworks` build phase diff in `MacDown 3000.xcodeproj/project.pbxproj`. CI regenerates these transiently on every run, so the committed files lag reality until this is done.

Related to #129

schuyler:
Addressed by #557.

`Tools/sign_sparkle.sh` re-signs Sparkle's five nested components inside-out with the release identity, `--options runtime`, and `--timestamp`; the app-bundle re-sign that follows re-seals over it. `Tools/verify_sparkle_signature.sh` asserts Developer ID authority, hardened runtime, team identifier equality, and a secure timestamp on every component at three gates — after signing, on the app inside the DMG, and at final pre-publish — all of which precede notarization.

[...]

Issue #129, opened 2025-11-20
Upgrade to Sparkle 2 with EdDSA signing and restore auto-updates


schuyler:
## Overview

MacDown's auto-update system is currently using obsolete Sparkle 1.18.1 with deprecated DSA signing. The appcast feeds are returning 503 errors, making auto-updates non-functional. With the new macdown.app domain and access to an Apple Developer account, we can restore auto-updates using modern Sparkle 2.

## Current State

**Sparkle Version:** 1.18.1 (via CocoaPods)
**Signing:** DSA (deprecated, scheduled for removal)
**Appcast URLs:**
- Stable: `https://macdown.app/sparkle/macdown3000/stable/appcast.xml` (503)
- Beta: `https://macdown.app/sparkle/macdown3000/testing/appcast.xml` (503)

**Code:** Using deprecated `SUUpdater` API in `MPMainController.m`

### 1. Upgrade Sparkle Framework
- Update `Podfile`: `pod 'Sparkle', '~> 2.0'`
- Run `bundle exec pod install`
- Minimum macOS requirement will increase to 10.11+

### 2. Migrate Code to Sparkle 2 API
- Replace `#import <Sparkle/SUUpdater.h>` with `#import <Sparkle/SPUStandardUpdaterController.h>`
- Update `MPMainController.m` to use new `SPUStandardUpdaterController` class
- Remove singleton pattern usage
- Update delegate methods for Sparkle 2

### 3. Generate EdDSA Signing Keys
- Use Sparkle's tooling to generate new EdDSA (ed25519) keypair
- Store private key securely (offline, NOT in repository)
- Add public key to Info.plist

### 4. Update Info.plist
- Remove: `SUPublicDSAKeyFile` and `dsa_pub.pem` reference
- Add: `SUPublicEDKey` with new EdDSA public key
- Update appcast URLs if needed

### 5. Remove DSA Key File
- Delete `MacDown/Resources/dsa_pub.pem`

### 6. Set Up Appcast Hosting
- Host appcast.xml files at macdown.app (HTTPS required)
- Use `generate_appcast` tool for creating signed appcasts
- Document release workflow

## Resources

- [Sparkle 2 Documentation](https://sparkle-project.org/documentation/)
- [Upgrading from Sparkle 1](https://sparkle-project.org/documentation/upgrading/)
- [EdDSA Migration Guide](https://sparkle-project.org/documentation/eddsa-migration/)

## Benefits

- Modern, maintained update framework
[...]

PR #556, opened 2026-08-04
Address issue #129: migrate to Sparkle 2 with EdDSA configuration

schuyler:
## Summary

Re-enables Sparkle auto-updates (disabled in place since #132) on Sparkle 2.9.5, migrating from the deprecated `SUUpdater` API to `SPUStandardUpdaterController`.

- `Podfile` re-enables `pod 'Sparkle', '~> 2.9'` (scoped to the `MacDown` target only); `Podfile.lock` and the CocoaPods-generated `[CP] Embed Pods Frameworks` build phase are committed.
- `MPMainController` creates and conditionally starts `SPUStandardUpdaterController` in `applicationDidFinishLaunching:`, guarded by `MPUpdaterDisabled()` (true under XCTest, or via a `-MPDisableUpdater YES` launch argument/persistent kill switch). Conforms to `NSMenuItemValidation` to gate the restored "Check for Updates…" menu item on `updater.canCheckForUpdates`.
- Beta/pre-release delivery uses Sparkle 2 channels rather than a dual feed-URL setup: a single `SUFeedURL`, with `allowedChannelsForUpdater:` returning `{"beta"}` or `{}` based on the existing `updateIncludesPreReleases` preference (unchanneled appcast items are always delivered; `SUBetaFeedURL` and `feedURLStringForUpdater:` are retired).
- `SUPublicEDKey` ships as a placeholder (base64 of 32 zero bytes) with an inline comment marking it inert; the old `SUPublicDSAKeyFile` plist entry and `MacDown/Resources/dsa_pub.pem` are removed.
- The "Include pre-releases" preferences checkbox is re-enabled (its force-disable code and "(unavailable)" xib title are removed).
- `plans/release-checklist.md` and `.claude/commands/release.md` document the one-time prerequisite: generating the real EdDSA keypair via Sparkle's vendored `generate_keys` tool and replacing the placeholder before a real release ships auto-updates. Appcast generation/hosting automation is a separate, not-yet-implemented follow-up.

## Related Issue

Related to #129

## Judgment Calls

Full requirements interpretation posted here: https://github.com/schuyler/macdown3000/issues/129#issuecomment-5164857744

- `MPUpdaterDisabled()` is a new mechanism (not itself one of the agreed decisions) needed to keep the updater from starting during tests/UI tests — checks for `XCTestCase` in-process, then a `-MPDisableUpdater` launch argument/default as a persistent kill switch.
- `-validateMenuItem:` is the first custom menu validation on this delegate; it gates only `checkForUpdates:` and falls through to `YES` for every other action, preserving prior always-enabled behavior for the rest of the menu.
- Sparkle 2.9 was chosen over the Podfile's originally-considered `~> 2.6`, and `use_frameworks!` is deliberately not set, since Sparkle ships as a prebuilt vendored binary framework rather than needing Swift module framework support.
- Sparkle's nested XPC components (Installer.xpc, Downloader.xpc) aren't individually re-signed by the CocoaPods embed script — that gap is explicitly owned by issue #553, not this PR.

## Testing

- TDD: `MacDownTests/MPUpdaterTests.m` (9 tests) written first against the not-yet-migrated code, then made to pass. Covers feed URL configuration, DSA/beta-feed retirement, EdDSA key well-formedness, updater controller class, the deferred-start-under-XCTest guard (asserted against the real `NSApp.delegate`, with explicit non-nil checks along the KVC chain), channel mapping in both directions, and menu wiring.
- `MacDownTests/MPPreferencesViewControllerResizabilityTests.m` gains a headless test confirming the pre-releases checkbox is no longer force-disabled.
- `MacDownTests/MPPreferencesTests.m` gains a round-trip test for `updateIncludesPreReleases`.
- Full suite run locally: 1194 tests executed, 0 unexpected failures (76 expected `XCTExpectFailure` cases, consistent with this project's baseline).


schuyler:
## Maintainer notes (self-review, not fixed in this PR)

Observed during self-review but deliberately left as-is:

- **ATS `NSAllowsArbitraryLoads = true`** is pre-existing (untouched by this PR) but is now exercised for the first time by a real outbound network path (update checks/downloads). Sparkle enforces its own HTTPS requirement for `SUFeedURL`, so this isn't a regression here, but it's a defense-in-depth gap this PR newly activates — worth a tracked follow-up if not already covered elsewhere.
- **`-MPDisableUpdater`'s NSUserDefaults branch of `MPUpdaterDisabled()` is structurally unreachable from XCTest** (the `XCTestCase`-loaded check always short-circuits first), so it's only exercised by `MacDownUITests.swift`'s launch argument — which sets the flag but asserts nothing about its effect. A broken key name or defaults lookup wouldn't fail any test; worst case is a live network call during a UI test run. Fixing this properly needs real UI-test assertions on menu-item state, which felt like more scope than this round warranted.
- **`project.pbxproj`'s diff contains substantial unrelated-entry reordering** (test file references for zoom/PDF-anchor/table tests etc. moved position with identical content) — a side effect of `pod install` resaving the file. Traced it fully; nothing dropped or altered, just noisy to read.
- **Thread-safety of `allowedChannelsForUpdater:`** reading `updateIncludesPreReleases` wasn't independently confirmed against Sparkle's documented threading contract for `SPUUpdaterDelegate` callbacks. No crash risk found either way; worth a one-line doc confirmation at some point.

Issue #558, opened 2026-08-04
Blank pipe-table header rows still render an empty <thead> line


macnotes:
## Problem

A pipe table whose header row is empty (e.g. `| | |`) still renders a visibly empty header bar in the preview:

```markdown
| | |
|---|---|
| **In** | Fri 30 Oct — VivaAerobus VB 7314... |
| **Out** | Sat 7 Nov — Aeroméxico AM 0640... |
```

Hoedown's table extension always wraps the first row of a pipe table in `<thead><tr><th>...</th></tr></thead>`, even when every `<th>` cell is empty. There's no post-processing anywhere in the rendering pipeline that detects and drops an all-empty header row, so the preview shows a blank header line above the data rows.

## Fix

Add a small Hoedown HTML renderer patch (a custom `table_header` callback) that inspects the rendered header row's content — after stripping HTML tags and entities — and, if nothing but whitespace remains, omits the `<thead>` element entirely instead of emitting an empty one. This only changes behavior when *every* header cell is empty; tables with any real header content render exactly as before.

Happy to open a PR for this.

schuyler:
thanks for the PR @macnotes !

PR #559, opened 2026-08-04
Address issue #558: omit empty <thead> from pipe tables

macnotes:
Related to #558

## Summary
- A pipe table whose header row is empty (e.g. `| | |`) still rendered a visibly empty `<thead>` bar, since Hoedown's table extension always wraps the first row in `<thead><tr><th>` regardless of cell content.
- Adds a Hoedown HTML renderer patch (`table_header` callback) that detects when every header cell is empty — after stripping tags/entities — and omits the `<thead>` element entirely in that case. Tables with any real header content render exactly as before.

## Test plan
- [x] Build succeeds (`xcodebuild ... -scheme MacDown`)
- [x] `MacDownTests/MPHTMLTabularizeTests` passes
- [ ] Manually verified in the preview: a table with an empty header row (`| | |`) no longer shows a blank header line; a table with real headers is unaffected

Note: `HGMarkdownHighlighterTests/testAllBundledThemesDefineHTMLForegroundColor` fails locally on plain `main` as well (unrelated to this change — appears to be a pre-existing environment-specific theme-resource-path issue on my machine), so it's not a regression introduced here.

### DIFF
--- a/MacDown/Code/Document/MPRenderer.m
+++ b/MacDown/Code/Document/MPRenderer.m
@@ -424,6 +424,7 @@ NS_INLINE hoedown_renderer *MPCreateHTMLRenderer(MPRenderer *renderer, int tocLe
     htmlRenderer->blockcode = hoedown_patch_render_blockcode;
     htmlRenderer->listitem = hoedown_patch_render_listitem;
     htmlRenderer->header = hoedown_patch_render_header;
+    htmlRenderer->table_header = hoedown_patch_render_table_header;
 
     hoedown_html_renderer_state_extra *extra =
         hoedown_malloc(sizeof(hoedown_html_renderer_state_extra));
--- a/MacDown/Code/Extension/hoedown_html_patch.c
+++ b/MacDown/Code/Extension/hoedown_html_patch.c
@@ -346,6 +346,58 @@ void hoedown_patch_render_header(
     hoedown_buffer_free(slug);
 }
 
+// Returns 1 if the rendered row content contains no visible text once HTML
+// tags and entities are stripped out (i.e. every cell in the row is empty).
+static int is_row_content_blank(const hoedown_buffer *content)
+{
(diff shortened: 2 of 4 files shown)


schuyler (changes requested):
@macnotes out of curiosity, why treat tags/entities as no-content? It seems at least possible that a user could put an image or a Unicode entity into a header.

schuyler:
Something we might do in a future branch: test coverage for the new callback, and the tag/entity-stripping walk that `is_row_content_blank()` shares with `slugify()`. We can file a follow-up issue for both once this merges, but please feel free to include that if you have the time.

The blank-detection point in the review above is the only outstanding item on this PR.

macnotes:
Good catch — fixed. \`is_row_content_blank()\` was stripping all tags and entities unconditionally, so an \`\`-only or \`&copy;\`-only header cell would have been misclassified as blank and its \`<thead>\` dropped.

Pushed two commits:
- Treat "replaced" elements (\`img\`, \`svg\`, \`video\`, \`audio\`, \`iframe\`, \`embed\`, \`object\`, \`canvas\`) as always-visible content, independent of surrounding text.
- Distinguish entities: whitespace-only ones (\`&nbsp;\`, \`&ensp;\`, \`&emsp;\`, \`&thinsp;\`, and their numeric refs) still count as blank; everything else (e.g. \`&copy;\`) counts as content.
- Added test coverage in \`MPRendererEdgeCaseTests.m\` for: blank header omits \`<thead>\`, image-only header keeps it, visible-entity-only header keeps it, whitespace-entity-only header still omits it, plus case-insensitive tag matching and a mixed blank/non-blank row.

Also swapped a \`strtol\` call on the (non-NUL-terminated) hoedown buffer for a manual bounded parse of numeric entity refs, since that was fragile even though safe today.

schuyler:
Thanks for this contribution @macnotes!

Issue #560, opened 2026-08-04
Preference to hide redundant horizontal rules in preview


macnotes:
## Problem

AI-generated markdown frequently inserts `---` horizontal-rule separators between sections. Most theme CSS already renders a border/rule under major headings (h1/h2) for visual separation, so the rendered result ends up with two lines back-to-back: the theme's heading border and the markdown `<hr>`. There's currently no way to suppress `<hr>` rendering without editing the markdown source itself.

## Fix

Add a "Hide horizontal rules" checkbox to the Rendering preferences pane (default on) that suppresses `<hr>` output in the preview/export via a new Hoedown renderer patch, without needing to touch the markdown source.

Happy to open a PR for this.

PR #561, opened 2026-08-04
Address issue #560: preference to hide horizontal rules in preview

macnotes:
Related to #560

## Summary
- AI-generated markdown frequently inserts `---` separators between sections. Most theme CSS already renders a border/rule under major headings for visual separation, so the rendered result ends up with two lines back-to-back.
- Adds a "Hide horizontal rules" checkbox to the Rendering preferences pane (default on) that suppresses `<hr>` output via a new Hoedown renderer patch, without needing to touch the markdown source.

## Test plan
- [x] Build succeeds (`xcodebuild ... -scheme MacDown`)
- [ ] Manually verified in the preview: with the preference on, `---` lines produce no `<hr>`; with it off, `<hr>` renders as before

### DIFF
--- a/MacDown/Code/Document/MPDocument.m
+++ b/MacDown/Code/Document/MPDocument.m
@@ -238,6 +238,8 @@ NS_INLINE NSColor *MPGetWebViewBackgroundColor(WebView *webview)
         flags |= HOEDOWN_HTML_HARD_WRAP;
     if (self.htmlCodeBlockAccessory == MPCodeBlockAccessoryCustom)
         flags |= HOEDOWN_HTML_BLOCKCODE_INFORMATION;
+    if (self.htmlHideHorizontalRules)
+        flags |= HOEDOWN_HTML_HIDE_HORIZONTAL_RULES;
     return flags;
 }
 @end
--- a/MacDown/Code/Document/MPRenderer.m
+++ b/MacDown/Code/Document/MPRenderer.m
@@ -424,6 +424,8 @@ NS_INLINE hoedown_renderer *MPCreateHTMLRenderer(MPRenderer *renderer, int tocLe
     htmlRenderer->blockcode = hoedown_patch_render_blockcode;
     htmlRenderer->listitem = hoedown_patch_render_listitem;
     htmlRenderer->header = hoedown_patch_render_header;
+    if (flags & HOEDOWN_HTML_HIDE_HORIZONTAL_RULES)
+        htmlRenderer->hrule = hoedown_patch_render_hrule_hidden;
 
     hoedown_html_renderer_state_extra *extra =
         hoedown_malloc(sizeof(hoedown_html_renderer_state_extra));
--- a/MacDown/Code/Extension/hoedown_html_patch.c
+++ b/MacDown/Code/Extension/hoedown_html_patch.c
@@ -346,6 +346,16 @@ void hoedown_patch_render_header(
     hoedown_buffer_free(slug);
 }
 
+// hrule replacement that emits nothing, used when the user has opted to hide
+// horizontal rules (e.g. redundant `---` section separators when headings
+// already provide visual separation via theme CSS).
+void hoedown_patch_render_hrule_hidden(
+    hoedown_buffer *ob, const hoedown_renderer_data *data)
+{
+    (void)ob;
+    (void)data;
+}
+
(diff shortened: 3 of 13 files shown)


schuyler (commented):
Thanks for the PR, @macnotes!

## Issues

- **Flag never set:** `HOEDOWN_HTML_HIDE_HORIZONTAL_RULES` is never OR-ed into `rendererFlags` — the composer at `MacDown/Code/Document/MPDocument.m:230-242` doesn't consult `htmlHideHorizontalRules`, so the new `hrule` callback is never installed and the checkbox has no effect on preview, export, or copy-as-HTML.
- **Coverage for the flag composition:** a test asserting `(preferences.rendererFlags & HOEDOWN_HTML_HIDE_HORIZONTAL_RULES) != 0` when the preference is on would catch this specific gap, which a renderer-level test passing the flag directly would not.
- **Default-on affects existing users:** the default lives in `loadDefaultUserDefaults`, which runs on every launch, so once the wiring is fixed every existing user's documents silently lose their `<hr>` output — worth calling out the intent explicitly, since this removes visible content rather than adding behavior.

## Suggestions

- Quick Look has its own flags composer at `MacDownCore/MPQuickLookPreferences.m:148-164` that won't pick up the preference, so previews in Finder would still show rules unless it's wired up too or the exclusion is stated.
- The other patched callbacks are installed unconditionally and test `state->flags` internally, so following that pattern here would keep the flag handling in one place.
- `MPMarkdownRenderingTests.m` already has `renderMarkdown:withExtensions:rendererFlags:` and a `testHorizontalRules` fixture, which makes rendering tests for `---`/`***`/`___`, rules inside blockquotes and lists, and a guard that `Title\n---` still renders `<h2>` fairly cheap to add.

---

macnotes:
Thanks for the thorough review — fixed all three issues:

**Flag never set:** \`MPPreferences (Hoedown)\`'s \`-rendererFlags\` (in \`MPDocument.m\`) now OR's in \`HOEDOWN_HTML_HIDE_HORIZONTAL_RULES\` when \`htmlHideHorizontalRules\` is on, matching the pattern of the other flags in that method. Also declared that category properly in \`MPPreferences.h\` (it previously had no header declaration at all, callable only via loose message-sends from within \`MPDocument.m\`) so it's callable — and testable — from elsewhere.

**Test coverage for flag composition:** Added \`testHideHorizontalRulesFlagComposition\` in \`MPPreferencesTests.m\`, asserting \`(preferences.rendererFlags & HOEDOWN_HTML_HIDE_HORIZONTAL_RULES)\` directly, which is exactly the gap a renderer-level test wouldn't have caught.

**Quick Look:** \`MPQuickLookPreferences\` gets a new \`-hideHorizontalRules\` reader (default YES, matching the app default) and \`MPQuickLookRenderer.m\` installs a local no-op \`hrule\` callback when it's set — mirroring the existing \`mp_quicklook_render_blockcode\`/\`_header\` pattern, since \`hoedown_html_patch.c\` isn't linked into \`MacDownCore\`.

Also added the suggested rendering tests: \`---\`/\`***\`/\`___\` hiding, rules inside blockquotes/lists (with assertions that surrounding content still renders), and a guard that a setext heading (\`Title\` underlined with \`---\`) still renders as \< h2 \> rather than being swallowed as a hidden rule.

Left the "install unconditionally, check \`state->flags\` internally" suggestion alone — Quick Look's renderer overrides are a separate, parallel implementation (no shared code with \`hoedown_html_patch.c\`), so there's no single place to centralize that check across both call sites

On the default-on note: issue #560 explicitly asked for "default on," so that's intentional — flagging in case it's worth a mention in release notes since it does change what existing users see in previews of documents using \`---\` as a separator.

macnotes:
Honestly, given the complexity of this PR, and perhaps its limited value to the average user, it won't hurt my feelings at all if you want to reject it.  

But for me it's a nice time-saver.

schuyler:
According to Copilot:

> The build is failing due to missing type definitions. The `hoedown_html_patch.h` file references `hoedown_renderer_data` and `hoedown_list_flags` types that aren't declared.
>
> Add these includes at the top of `hoedown_html_patch.h` after the `#define` guard:

```c
#include "hoedown/document.h"
#include "hoedown/renderer.h"
```

Sorry for all the back and forth. If I could edit this branch, I'd just go ahead and make the change and push again

Issue #478, opened 2026-06-30
File watcher: guard resource watchers, handle transient paths, fix stale watcher leak


schuyler:
## Context

PR #424 (skip file watchers on remote volumes, related to #371) addressed document file watching but left three gaps identified during review:

### Issues

1. **Resource watchers unguarded:** `MPResourceWatcherSet -addWatcherForPath:` is a second `MPFileWatcher` instantiation site with no `canWatchPath:` check, so resource files (images/CSS/scripts) on remote volumes still hit the same failure #424 fixes for documents.

2. **Non-existent path regression:** `NSURLVolumeIsLocalKey` requires the file to exist, so `canWatchPath:` returns NO for a path that transiently doesn't exist (e.g. after an external editor's atomic rename-replace save), which can leave auto-reload disabled. Falling back to the parent directory for the locality check would avoid this.

3. **Stale watcher on early-return:** The new early-return in `MPDocument -startFileWatching` runs before `stopFileWatching`, so a previously-armed watcher can leak if `canWatchPath:` returns NO for an already-watched session.

### Additional improvements

- **Branch coverage:** the empty-string and non-existent-path branches of `canWatchPath:` are untested.
- **Diagnostics:** pass `NSError *` instead of `NULL` to `getResourceValue:forKey:error:` and log it for easier diagnosis of "Device not configured" failures.
- **Doc comment:** the header comment describes only the local-volume case, though the method also returns NO for nil, empty, and non-existent paths.

From review of PR #424: https://github.com/schuyler/macdown3000/pull/424#pullrequestreview-4542746014

Related to #371

schuyler:
## Requirements & decisions

Picking this up as an rc.2 blocker. After reading the merged #424 code, here's the scope and a couple of corrections to the framing in the issue body:

**Gap 2 — non-existent-path regression (the real bug).** Confirmed. `+[MPFileWatcher canWatchPath:]` reads `NSURLVolumeIsLocalKey`, which requires the item to exist; during an external editor's atomic rename-replace save the file transiently vanishes, so the check returns `NO` and auto-reload silently dies. **Fix:** when the file doesn't exist, probe the **parent directory's** volume instead; return `NO` only if neither path nor parent resolves, or the volume is non-local.

[...]

PR #568, opened 2026-08-12
Fix Editor preferences pane hang in wrapping-checkbox layout test

schuyler:
## Summary

- The Editor preferences pane's "Behavior" `NSBox` positioned its checkboxes relative to the box's *content view* instead of the box itself, unlike every other box in the preferences panes (General, Markdown). The content view had no constraints of its own tying it to the box, so the box's height had no required-priority path back to its checkboxes' actual size.
- `+[MPPreferencesViewController addHeightConstraintsForWrappingCheckboxesInView:]` grows checkboxes past their default height when their titles wrap, and in that state `layoutSubtreeIfNeeded` never converges for this box. `MacDownTests/MPPreferencesViewControllerResizabilityTests.m`'s `testWrappingCheckboxHeightsAccommodateMultiLineText` exercises exactly this by forcing wrapped titles, and hangs CI indefinitely as a result (observed on PR #493, unrelated to that PR's diff — this bug is pre-existing on `main`).
- The four affected constraints in `MPEditorPreferencesViewController.xib` now reference the box instead of the content view, matching the General and Markdown panes' boxes. The constants are adjusted for the box's chrome inset (1pt left/right/bottom, 15pt top), so the checkboxes' resolved position and the box's total height are unchanged — verified algebraically against the box/content-view frame deltas, not just asserted.
- Also adds a 30-minute job/step timeout and `-test-timeouts-enabled YES -maximum-test-execution-time-allowance 120` to the `xcodebuild test` invocations in `test.yml` and `release.yml`'s "Run tests before release" step, which lacked them — matching the existing convention already used in `smoke-test.yml` and `release.yml`'s UI-test step. This bounds any future test hang to minutes instead of GitHub's 6-hour default, independent of whether this particular fix is complete.

## Verification

- Root cause and fix were each independently reviewed by a separate agent (design review before implementation, full-branch implementation review after) per this repo's review process.
- XML well-formedness and constraint-graph completeness verified programmatically.
- Geometry preservation (checkbox position, box height) verified algebraically from the box/content-view frame deltas — not run, since this environment has no macOS/Xcode.
- **Not yet visually verified on macOS.** The math says nothing should move, but a human should eyeball the Editor preferences pane (English and one wrapping locale) before merge.

## Test plan

- [ ] CI passes on both `macos-14` and `macos-26` without hanging
- [ ] Manually open Preferences → Editor on macOS and confirm the "Behavior" box's checkboxes are laid out correctly (no clipping, no unexpected shift)
- [ ] Same check with a long-title locale (e.g. French) if convenient, since that's the scenario the checkbox-wrapping code exists for

---
_Generated by [Claude Code](https://claude.ai/code/session_01Dxfe24SHv2odFfQ4C37HFP)_

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -168,12 +168,15 @@ jobs:
         uses: ./.github/actions/setup-macdown
 
       - name: Run tests before release
+        timeout-minutes: 30
         run: |
           echo "🧪 Running unit tests to ensure code quality before release..."
           xcodebuild test \
             -workspace "MacDown 3000.xcworkspace" \
             -scheme MacDown \
-            -derivedDataPath DerivedData
+            -derivedDataPath DerivedData \
+            -test-timeouts-enabled YES \
+            -maximum-test-execution-time-allowance 120
           echo "✅ All tests passed"
 
       # Keep in sync with the launch-test job in smoke-test.yml
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -13,6 +13,7 @@ jobs:
         os: [macos-14, macos-26]
       fail-fast: false
     runs-on: ${{ matrix.os }}
+    timeout-minutes: 30
     permissions:
       contents: read
 
@@ -31,7 +32,9 @@ jobs:
         xcodebuild test \
           -workspace "MacDown 3000.xcworkspace" \
           -scheme MacDown \
-          -derivedDataPath DerivedData
+          -derivedDataPath DerivedData \
+          -test-timeouts-enabled YES \
+          -maximum-test-execution-time-allowance 120
 
     - name: Upload test results
       if: failure()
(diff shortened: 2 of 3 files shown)


schuyler:
CI update: the fix works — `Run Unit Tests (macos-14)` (the job that hung on PR #493) now completes in ~2 minutes with no hang, and `Verify Release Build (Dry Run)` passed.

`Run Unit Tests (macos-26)` failed, but not from a hang or timeout — it ran the full 397-test suite normally and failed on one test: `-[MPResourceWatcherSetTests testDelegateCalledOnAtomicSave]`. That test is a wall-clock race (`dispatch_after` 0.1s, then `waitForExpectationsWithTimeout:2.0`) in the file-watching code (`MPResourceWatcherSet`), last touched by an unrelated commit (`2f62a58`, issue #478) — nothing this PR changes. It's the kind of test that can flake under CI load rather than a real regression.

PR #569, opened 2026-08-12
Rework issue #451: draggable table column resizing, made ephemeral

schuyler:
Draggable table column resizing in the live preview, reworked from @aseelye's PR #451 contribution to hold width state in memory only, with no document mutation or persistence of any kind.

## Behavior

- Drag a column-boundary handle in a rendered table to resize it live in the preview.
- Double-click a handle to reset that column to natural sizing.
- Widths survive incremental preview updates (typing) because the preview's JS context persists across DOM replacement.
- Widths reset on a full preview reload (style change, window reopen) and on document close. Nothing is written to the document or to disk.
- Export (HTML/PDF) is unaffected by resize state.

## What changed from @aseelye's original PR #451

His interactive front end (drag handles, table identity via header-hash + index, minimum column width) is unchanged. Removed entirely: the native persistence layer — bridge token, `x-macdown-table-layout://` custom URL scheme, `table-layouts.json` sidecar file, and the associated load/save/reset logic in `MPDocument.m`/`MPRenderer.m`/`.h`. `table-resize.js` now holds width state in a closure-scoped map instead of round-tripping through the native bridge.

His authorship is preserved via a straight cherry-pick (`a5f9b40`) as the base commit, with the persistence removal as a separate commit on top.

## Judgment calls

- **No Markdown parsing, no export of resized widths.** The feature is purely DOM-level and is unaffected by the hoedown → cmark-gfm parser swap (#515 / v3000.1).
- **Table identity is header-hash + index**, unchanged from the original PR. Renaming a header or reordering tables (which shifts the index) loses that table's widths for the session — accepted as an ephemeral-view-state tradeoff.
- **`export.css`'s bundled `th { white-space: nowrap }` change rides along unchanged** from the original PR. It affects all table rendering, including export, not just the live-resize feature — a UX change serving the same table-readability goal, kept as part of this PR rather than split out.
- **In-document persistence remains a future feature**, tracked separately against a dash-count design built for cmark-gfm rather than the sidecar-file approach removed here.

## Testing

`CI=true xcodebuild test -workspace "MacDown 3000.xcworkspace" -scheme "MacDown (MacDown 3000 project)" -destination 'platform=macOS'`: 1225 tests, 76 expected failures, 0 unexpected.

Headless tests cover: the renderer emits no layout JSON or table-layout bridge token; `table-resize.js` is registered in the preview scripts; export HTML/CSS behavior is unaffected; the removed URL-scheme handler leaves no trace. Drag interaction and width re-application after DOM replacement run only in a live WebView and are manual-QA items.

Related to #451


schuyler:
Local test run against this branch (three separate invocations — warm DerivedData, fully cold DerivedData, and reproducing CI's exact scheme name and xcodebuild flags) passes 1303 tests: 76 expected `XCTExpectFailure` failures, 0 unexpected.

The `Tests` workflow failures above (macos-14 cancelled, macos-26 failed) are the intermittent CI-runner hang under separate investigation — both matrix jobs stall for ~2 minutes on unrelated tests with no local reproduction across three attempts, consistent with a runner-level issue rather than a regression in this branch.

